PR #1, opened 2024-10-07
00 setup

riraum:
Requesting review.

### DIFF
--- a/00_setup/main.go
+++ b/00_setup/main.go
@@ -30,6 +30,7 @@ func main() {
 	fmt.Println("Hello", "World!", 42, 3.14, true, false, []int{1, 2, 3})
 
 	// TODO: Print "Hello, mom!"
-
+	fmt.Println("Hello, mom!")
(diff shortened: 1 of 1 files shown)


00_setup/main.go
```diff
@@ -31,6 +30,7 @@ fmt.     Println(      "Hello, World!"     )
 	fmt.Println("Hello", "World!", 42, 3.14, true, false, []int{1, 2, 3})
 
 	// TODO: Print "Hello, mom!"
-
+	fmt.Println("Hello, mom!")
 	// TODO: Print a couple of numbers
+	fmt.Println("333, 3.1773")
```
nobe4: No, this prints a single string which contains numbers.
I meant: print a couple of numbers.

nobe4 (changes requested):

00_setup/main.go
riraum: That makes sense! Thanks.

nobe4 (approved):

PR #5, opened 2024-10-07
04 boolean

riraum:
Requesting review.

### DIFF
--- a/04_boolean/main.go
+++ b/04_boolean/main.go
@@ -74,15 +74,24 @@ func main() {
 	fmt.Println(true, false)
 
 	// TODO: Define a variable named `awesome` and set it to `true`.
+	var awesome bool = true
 
 	// TODO: Change the value of `awesome` to `false`.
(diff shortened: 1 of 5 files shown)


04_boolean/main.go
```diff
@@ -104,9 +126,34 @@ func main() {
...
 	// 2. Use nested if-else statements
 	// TODO: Try both ways.
+	if firstName == "John" {
+		if lastName == "Doe" {
+			fmt.Println("You are John Doe")
+		} else {
+			fmt.Println("You are another John")
+		}
+	}
+	if firstName != "John" {
```
nobe4: You should here use `else`.
Since the `if` before that is checking the opposite condition.

E.g.

```
if A { ... }
if !A { ... }

should be

if A { ... }
else { ... }
```

riraum: Took me a while, but makes sense!

nobe4 (approved):

PR #7, opened 2024-10-07
05 loops

riraum:
Requesting review.
I'm not sure about the Fibonacci loop, will review it once more later.
The Fizzbuzz loop I only have the non nested version running for now.
Can also just wait until I gave both another try.

### DIFF
--- a/05_loops/main.go
+++ b/05_loops/main.go
@@ -73,18 +73,28 @@ func main() {
 
 	// TODO: Write a loop that print the numbers from 0 to 100, inclusive, but
 	// only the even numbers.
-	var l int = 0
-	for {
-		if l > 100{
-			break
-		}
-		if l % 2 == 0 {
-			fmt.Println(l)
-			}
-		l += 1
-	}
+	// var l int = 0
+	// for {
+	// 	if l > 100{
+	// 		break
+	// 	}
+	// 	if l % 2 == 0 {
+	// 		fmt.Println(l)
+	// 		}
+	// 	l += 1
+	// }
 
 	// TODO: Write a loop that prints the first 100 odd numbers.
+	// var m int = 0
+	// for {
+	// 	if m > 100 {
+	// 		break
+	// 	}
+	// 	if m % 2 != 0 {
+	// 		fmt.Println(m)
+	// 	}
+	// 	m += 1
+	// }
 
 	// TODO: Define a variable and write a loop that calculates the sum of the
 	// numbers from 0 to 100, inclusive. Print the result.
--- a/05_loops/main.go
+++ b/05_loops/main.go
@@ -3,7 +3,7 @@ package main
 import "fmt"
 
 /*
-
+TEST
 Go has loops.
 
 Loops are one of the main building blocks of any programming language.
--- a/05_loops/main.go
+++ b/05_loops/main.go
@@ -85,16 +85,23 @@ func main() {
 	// }
 
 	// TODO: Write a loop that prints the first 100 odd numbers.
-	// var m int = 0
-	// for {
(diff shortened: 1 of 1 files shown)


05_loops/main.go
```diff
@@ -43,28 +42,130 @@ it's also the most explicit.
...
+	// var m int = 0
+	// for {
+	// 	if m > 100 {
+	// 		break
+	// 	}
+	// 	if m % 2 != 0 {
+	// 		fmt.Println(m)
+	// 	}
+	// 	m += 1
+	// }
```
nobe4: This is incorrect.

This prints the odd numbers from 0 to 100.

What I asked is print the _first 100 odd numbers_.

E.g. 
- if I asked "print the odd numbers from 1 to 5", then I expect "1, 3, 5"
- if I asked "print the first 5 odd numbers, then I expect "1, 3, 5, 7, 9"

05_loops/main.go
```diff
@@ -43,28 +42,130 @@ it's also the most explicit.
...
 	// TODO: Write a fizzbuzz.
+	// Fizz buzz is a group word game for children to teach them about division.
+	//  Players take turns to count incrementally, replacing
+	// any number divisible by three with the word "fizz", and any number divisible by five
+	//  with the word "buzz", and any number divisible
+	// by both three and five with the word "fizzbuzz".
+	// If none applies, return value.
+	// var o int = 1
+	// var p string = "fizz"
+	// var q string = "buzz"
```
nobe4: While this works, those variable names are confusing. I would change that.

05_loops/main.go
```diff
@@ -43,28 +42,130 @@ it's also the most explicit.
...
+	// 		}
+	// 		if o%5 == 0 {
+	// 			fmt.Println(q)
+	// 	} else {
+	// 		fmt.Println(o)
+	// 	}
+	// 	o += 1
+	// }
 
 	// TODO: Write a loop that prints the first 100 Fibonacci numbers.
```
nobe4: I just realized that `100` is too big, make it 20 instead.
```suggestion
	// TODO: Write a loop that prints the first 20 Fibonacci numbers.
```

05_loops/main.go
```diff
@@ -43,28 +42,130 @@ it's also the most explicit.
...
+	// Fibonacci sequence is a sequence in which each number is the sum of the two preceding ones
+	var r int = 0
+	var s int = 1
+	var t int = 0
+	for {
+		if t > 100 {
```
nobe4:
```suggestion
		if t > 20 {
```

nobe4 (changes requested):
Added a couple of comments

05_loops/main.go
riraum: This one is the reason why I didn't finish all your suggested changes yesterday. But after a while of researching, I came to the solution by contemplating.

05_loops/main.go
riraum: You mean because p and q look so similar? Gave them some more distinct names as I think I ran out of individual letters of the alphabet.

riraum:
Fixed the nested fizzbuzz loop. It did make more sense today, thinking and looking at 04 booleans loop helped.

05_loops/main.go
```diff
@@ -53,18 +52,130 @@ func main() {
...
+	for {
+		if m%2 != 0 {
+			fmt.Println(m)
+			z += 1
+		}
+		m += 1
+		if z > 100 {
+			break
+		}
+	}
```
nobe4, line 104: Yes! That what I had in mind!

nobe4 (approved):

PR #8, opened 2024-10-08
06 functions

### LINKED ISSUE #9: clarify scope and variable locality
cc https://github.com/riraum/forward/pull/8#discussion_r1793073613

riraum:
Please review.

### DIFF
--- a/06_functions/main.go
+++ b/06_functions/main.go
@@ -101,8 +101,7 @@ func main() {
 // TODO: Write a function that takes two integers as parameters and returns
 // the difference of the two integers.
 func diff(num1 int, num2 int) int {
-	var diff int = num2 - num1
-	return diff
+	return num2 - num1
 }
 
 // TODO: Write a function that takes a string and an integer as parameters.
--- a/06_functions/main.go
+++ b/06_functions/main.go
@@ -100,8 +100,8 @@ func main() {
 
 // TODO: Write a function that takes two integers as parameters and returns
 // the difference of the two integers.
-func diff(a int, b int) int {
-	var diff int = b - a
+func diff(num1 int, num2 int) int {
+	var diff int = num2 - num1
 	return diff
 }
 
@@ -123,58 +123,57 @@ func printInput(str string, times int) {
 // many times as the integer and return the result.
 // Call the function a couple of times with different values and print the
 // results.
-func concatString(str2 string, times2 int) string {
-	var count2 = 1
-	// var result2 = ""
+func concatString(str string, times int) string {
+	var count = 1
 	for {
-		if count2 > times2 {
+		if count > times {
 			break
 		}
-		str2 += str2
-		count2++
+		str += str
+		count++
 	}
-	return str2
+	return str
 }
 
 // TODO: Write a function that takes a single integer as a parameter and
 // prints all numbers from 1 to that number.
 func printToInt(parInt int) {
-	var count3 = 1
+	var count = 1
 	for {
-		if count3 > parInt {
+		if count > parInt {
 			break
 		}
-		fmt.Println(count3)
-		count3++
+		fmt.Println(count)
+		count++
 	}
 }
 
 // TODO: Write a function that takes 2 strings as parameters and returns
 // the sentence "Hello <name1>, and <name2>."
 // Prints the result.
-func returnStrings(str3 string, str4 string) string {
-	var result = "Hello " + str3 + ", and " + str4 + "."
+func returnStrings(str1 string, str2 string) string {
+	var result = "Hello " + str1 + ", and " + str2 + "."
 	return result
 }
 
 // TODO: Write a function that takes a string as a parameter and prints:
 // "The string '<the string>' has <number of characters> characters."
-func length(str5 string) {
-	var length int = len(str5)
-	fmt.Println("The string", str5, "has", length, "characters.")
+func length(str string) {
+	var length int = len(str)
+	fmt.Println("The string", str, "has", length, "characters.")
 }
 
 // TODO: Update the previous function to handle the cases where the string
 // is empty or has a single character.
-func lengthAdvanced(str6 string) {
-	var length = len(str6)
+func lengthAdvanced(str string) {
+	var length = len(str)
 	// TODO Put generic print text in variable
-	if str6 == "" {
+	if str == "" {
 		fmt.Println("Empty string.")
 	} else if length == 1 {
-		fmt.Println("The string", str6, "has", length, "character.")
+		fmt.Println("The string", str, "has", length, "character.")
 	} else {
-		fmt.Println("The string", str6, "has", length, "characters.")
+		fmt.Println("The string", str, "has", length, "characters.")
 	}
 }
 
@@ -184,21 +183,19 @@ func lengthAdvanced(str6 string) {
 // Using those functions, write a function that takes two integers and
 // returns the result of the following formula: a * b + a
 // Print the result.
-func mul(number1 int, number2 int) int {
-	var prod int = number1 * number2
+func mul(num1 int, num2 int) int {
+	var prod int = num1 * num2
 	return prod
 }
 
-func add(number1 int, number2 int) int {
-	var sum int = number1 + number2
+func add(num1 int, num2 int) int {
+	var sum int = num1 + num2
 	return sum
 }
 
-func doingMath(number1 int, number2 int) int {
-	var mulValue = mul(number1, number2)
-	// fmt.Println(mulValue)
-	var addValue int = add(mulValue, number1)
-	// fmt.Println(addValue)
+func doingMath(num1 int, num2 int) int {
+	var mulValue = mul(num1, num2)
+	var addValue int = add(mulValue, num1)
 	return addValue
 }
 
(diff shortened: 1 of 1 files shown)


06_functions/main.go
```diff
@@ -80,6 +79,13 @@ func main() {
 
 	// TODO: Write a function that takes two integers as parameters and returns
 	// the difference of the two integers.
+	func diff(a int, b int) int {
+		let sum int = b - a
+		return sum
+	}
+	diff(1, 9)
+	diff(9, 1)
+	diff(4, -10)
```
riraum: I'm getting a `./main.go:82:7: syntax error: unexpected name diff, expected (
./main.go:83:7: syntax error: unexpected name sum at end of statement`
when I try either to `fmt` or `run` it.
Researched, but not sure what the exact issue is so far.

riraum: Fixed. Apparently I needed to close the main function before creating other functions.

06_functions/main.go
```diff
@@ -74,42 +73,64 @@ Note: functions can be defined in any order in the file.
...
+// For each exercise, call the functions a couple of times with different
+// values and print the results.
 
-	// TODO: Write a function that takes two integers as parameters and returns
-	// the difference of the two integers.
+// TODO: Write a function that takes two integers as parameters and returns
+// the difference of the two integers.
+func diff(a int, b int) int {
+	var sum int = b - a
+	return sum
```
nobe4: Just a name suggestion

```suggestion
	var diff int = b - a
	return diff
```

riraum: I tried to directly return actually, without variable declaration, but didn't find a compliant way.
You don't see an issue with the variable having the same name as the function, as it's inside the function I guess, okay sure.
I was thinking about converting the int to a string for example.

06_functions/main.go
```diff
@@ -74,42 +73,67 @@ Note: functions can be defined in any order in the file.
...
-	// the difference of the two integers.
+// TODO: Write a function that takes a string and an integer as parameters.
+// The function should print the string as many times as the integer.
+func printInput(c string, d int) {
+	var result = ""
+	var count = 1
+	for {
+		if count > d {
+			break
+		}
```
nobe4: @riraum ping, can you see me?

06_functions/main.go
```diff
@@ -74,42 +73,67 @@ Note: functions can be defined in any order in the file.
...
+func diff(a int, b int) int {
+	var sum int = b - a
+	return sum
+}
 
-	// TODO: Write a function that takes two integers as parameters and returns
-	// the difference of the two integers.
+// TODO: Write a function that takes a string and an integer as parameters.
+// The function should print the string as many times as the integer.
+func printInput(c string, d int) {
```
nobe4: Using good naming for parameters is crucial.
It's a bit subjective but using 1 letter is usually frown upon.
How about:
```suggestion
func printInput(str string, count int) {
```

riraum: That sounds good, thanks!

06_functions/main.go
```diff
@@ -74,41 +73,134 @@ Note: functions can be defined in any order in the file.
...
+	var length = len(str6)
+	// TODO Put generic print text in variable
+	if str6 == "" {
+		fmt.Println("Empty string.")
+	} else if length == 1 {
+		fmt.Println("The string", str6, "has", length, "character.")
+	} else {
+		fmt.Println("The string", str6, "has", length, "characters.")
+	}
+}
```
riraum: @nobe4 I tried creating a variable for `"The string", str6, "has", length` at least, to be more DRY.
So far wasn't successful to workaround the type mismatch. Did delve a bit into Go documentation regarding that. 
Not sure if it's worth pursuing at the moment or not.

riraum:
Ready for review @nobe4

06_functions/main.go
```diff
@@ -74,41 +73,134 @@ Note: functions can be defined in any order in the file.
...
+	var result = ""
+	var count = 1
+	for {
+		if count > times {
+			break
+		}
+		fmt.Println(str)
+		count++
+	}
+	fmt.Println(result)
```
nobe4: You are not using the variable `result` here.

06_functions/main.go
```diff
@@ -74,41 +73,134 @@ Note: functions can be defined in any order in the file.
...
-	// Call the function a couple of times with different values and print the
-	// results.
+// TODO: Write a function that takes a string and an integer as parameters.
+// The function should concatenate (i.e. join) the string with itself as
+// many times as the integer and return the result.
+// Call the function a couple of times with different values and print the
+// results.
+func concatString(str2 string, times2 int) string {
+	var count2 = 1
+	// var result2 = ""
```
nobe4:
```suggestion
```

06_functions/main.go
```diff
@@ -74,41 +73,134 @@ Note: functions can be defined in any order in the file.
...
+	for {
+		if count2 > times2 {
+			break
+		}
+		str2 += str2
+		count2++
+	}
+	return str2
+}
```
nobe4: There's a way to write that without an additional counter.
Can you figure out how?

nobe4 (commented):
Minor comments, otherwise looks good!

06_functions/main.go
riraum: Can you clarify if you mean no counter at all?
Because there's just one counter in the function. 
I tried to make my variable naming clearer in the latest commit.

06_functions/main.go
```diff
@@ -74,41 +73,130 @@ Note: functions can be defined in any order in the file.
...
+	printToInt(6)
+	fmt.Println(returnStrings("Anna", "Hanna"))
+	length("Gopher")
+	length("Go rocks")
+	lengthAdvanced("A")
+	lengthAdvanced("")
+	lengthAdvanced("Luna")
+	fmt.Println(doingMath(3, 4))
+	fmt.Println(doingMath(5, 10))
+}
```
riraum, line 99: How can I use declare and call functions within the main function?
I had the error `syntax error: unexpected name functionName, expected (` before I closed the main function before I declared new functions outside of it and calling them within the main function.
It does cause fmt to adjust the spacing and cause these huge diffs though.

nobe4 (approved):

riraum (commented):
From my side seems complete. Left some comments with questions.

06_functions/main.go
nobe4: Yup, you can't add strings to numbers with `+`. However, there are ways to create such variables, which we'll see later :)

06_functions/main.go
nobe4: In Go you cannot define a function inside another function: https://go.dev/play/p/kRlOfhC1sii

The only exception is when the function is a parameter, but we'll get to that later.

06_functions/main.go
nobe4: Try writing the same function without using `count2`.

Hint:
```go
func concatString(str2 string, times2 int) string {
	for {
		if ??? > 0 {
			break
		}
		str2 += str2

		???
	}
	return str2
}
```

06_functions/main.go
nobe4: conversion between types is usually a bad idea. Especially if you have the same entry/exit types, don't try converting things to strings.

Does this not work?

```go
func diff(a int, b int) int {
  return a - b
}
```

riraum: Somehow it works. Not sure what I did before.

06_functions/main.go
riraum: Okay, good. Then I can/have to close the main function, before I start writing other functions.

06_functions/main.go
```diff
@@ -123,13 +123,13 @@ func printInput(str string, times int) {
...
-		if count > times {
+		if times < 1 {
 			break
 		}
 		str += str
-		count++
+		times--
+		fmt.Println(times)
 	}
 	return str
```
riraum: @nobe4 Tried your suggestion. But now it causes to string to increase too much.
You are certain about `< 0` right? I will give it a try later again.

nobe4: maybe `< 0` is too much, but the idea is to decrement `times` as you found out :)

06_functions/main.go
riraum: @nobe4 Revised. If the concatenated string result should be:
> times = 4
> string = Rice
> concatString = RiceRiceRiceRiceRice

Then I have to think more on how to get that result.

nobe4: True, then maybe this: 
```go
func concatString(str2 string, times2 int) string {
        var result string = ""
	for {
		if ??? > 0 {
			break
		}
		result += str2

		???
	}
	return result
}
```

riraum: I did think of declaring another string too, that seems the only way I can think of at least.
Apart from a counter.

riraum: How about now? If you want `> 0` in the logic, I would give it another go.

nobe4: I think the code looks good now, as long as it prints the correct result :)

PR #10, opened 2024-10-09
07 boolean 2

riraum:
Ready


07_boolean_2/main.go
```diff
@@ -15,56 +15,136 @@ E.g.:
...
+//
+// - If the string is shorter than the integer, return "shorter"
+func strIntCheck(str string, num int) {
+	// var stringLength = len(str)
+	// var integerLength = len(num)
+	if str == "" {
+		fmt.Println("empty")
+		// TODO find way to get int length
+		// } else if len(str) > len(num) {
+		if len(str)%2 == 0 {
```
riraum: @nobe4 Unsure currently how to calculate length of `int`.
At least without anther import.
Maybe some loop?

07_boolean_2/main.go
```diff
@@ -15,56 +15,136 @@ E.g.:
...
+		// If average grade  >= 90
+	} else if avg >= 90 {
+		if grade1 < 80 && grade2 < 80 && grade3 < 80 && grade4 < 80 && grade5 < 80 {
+			fmt.Println("A+")
+		} else if grade1 < 70 && grade2 < 70 && grade3 < 70 && grade4 < 70 && grade5 < 70 {
+			fmt.Println("A")
+		} else if grade1 < 70 || grade2 < 70 || grade3 < 70 || grade4 < 70 || grade5 < 70 {
+			fmt.Println("A-")
+		}
+	}
```
riraum: @nobe4 WAY too messy. So far didn't figure out yet how to make this more DRY.

07_boolean_2/main.go
```diff
@@ -15,56 +15,136 @@ E.g.:
...
+		fmt.Println("NOR", true)
+	} else {
+		fmt.Println("NOR", false)
+	}
+	// XOR
+	if bool1 == true && bool2 == false || bool2 == true && bool1 {
+		fmt.Println("XOR", true)
+	} else if bool1 == bool2 || bool2 == bool1 {
+		fmt.Println("XOR", false)
+	}
```
nobe4: No, use logical operators.

```suggestion
	fmt.Println("NAND", ???)
	fmt.Println("NOR", ???)
	// if you really want to do an extra step, use a variable.
	var xor bool = ???
	fmt.Println("XOR", xor)
```

07_boolean_2/main.go
```diff
@@ -15,56 +15,136 @@ E.g.:
...
-	// the following checks:
-	// - If the string is empty, return "empty"
-	// - If the string is longer than the integer
-	//   - If the integer is even, return "longer and even"
-	//   - If the integer is odd, return "longer and odd"
-	// - If the string is shorter than the integer, return "shorter"
+// TODO: Write a function that takes a string and an integer and performs
+// the following checks:
+// - If the string is empty, return "empty"
+// - If the string is longer than the integer
```
nobe4:
```suggestion
// - If the string is longer than the value of the integer
```

07_boolean_2/main.go
nobe4: does https://github.com/riraum/forward/pull/10/files#r1793296769 clarify the exercise?

07_boolean_2/main.go
```diff
@@ -15,56 +15,136 @@ E.g.:
...
+// "C"
+//   - If no grade is below 70, return "C+"
+//   - If no grade is below 60, return "C"
+//   - If any grade is below 60, return "C-"
+func calcGrade(grade1 int, grade2 int, grade3 int, grade4 int, grade5 int) {
+	var avg = (grade1 + grade2 + grade3 + grade4 + grade5) / 5
+	// var allGradesAnd = grade1 && grade2 && grade3 && grade4 && grade5
+	// var allGradesOr = grade1 || grade2 || grade3 || grade4 || grade5
+	// if grade.. > 100 || grade.. < 0 print invalid
+	if (grade1 < 0 || grade1 > 100) || (grade2 < 0 || grade2 > 100) || (grade3 < 0 || grade3 > 100) || (grade4 < 0 || grade4 > 100) || (grade5 < 0 || grade5 > 100) {
```
nobe4: how about you create a function called `gradeValid`?

07_boolean_2/main.go
```diff
@@ -15,56 +15,136 @@ E.g.:
...
+		fmt.Println(("invalid grade"))
+		// If average grade  >= 90
+	} else if avg >= 90 {
+		if grade1 < 80 && grade2 < 80 && grade3 < 80 && grade4 < 80 && grade5 < 80 {
```
nobe4: how about you create a function called `anyValueUnder(...)` that returns true/false?

07_boolean_2/main.go
nobe4: Added some suggestions.

07_boolean_2/main.go
nobe4: maybe also `allValuesOver(...)`?

07_boolean_2/main.go
riraum: YES. I do appreciate the challenge, hahaha.

07_boolean_2/main.go
```diff
@@ -15,56 +15,129 @@ E.g.:
...
+	// if bool1 == true && bool2 == true {
+	// 	fmpt.Println("true")
+	// }
+	// Direct print statements
+	// AND // true, true == true || true, false == false || false, false == false || false, true == false
+	fmt.Println("AND", bool1 && bool2)
+	// OR // true, true == true || true, false == true || false, false == false || false, true == true
+	fmt.Println("OR", bool1 || bool2)
+	// NAND // true, true == false || true, false == true, || false, false == true || false, true == true
+	fmt.Println("NAND", bool1 != bool2 || !bool1)
```
riraum: @nobe4 I had `bool1 != bool2 || bool1 == false || bool2 == false)` and that made sense to me.

But I got told `should omit comparison to bool constant, can be simplified to !bool1 (S1002)go-staticcheck`. So I did.
I don't think I have fully internalized the logic.

nobe4: `bool1 == false` is strictly equivalent to `!bool1`.

`bool1 == true` is strictly equivalent to `bool1`.

07_boolean_2/main.go
```diff
@@ -15,56 +15,129 @@ E.g.:
...
+	//  NOR // true, true ==  false || true, false == false || false, false == true || false, true == false
+	fmt.Println("NOR", !bool1 && !bool2)
```
nobe4: X NOR Y = NOT(X OR Y)

What you wrote is (NOT X) AND (NOT Y).
It is technically the same but is harder to read.

07_boolean_2/main.go
```diff
@@ -15,56 +15,129 @@ E.g.:
...
+	// // XOR // true, true == false || true, false == true || false, false == false || false, true == true
+	fmt.Println("XOR", bool1 != bool2)
```
nobe4: This is correct, but try making one without `!=`.

07_boolean_2/main.go
nobe4: This is incorrect.

X NAND Y = NOT(X AND Y)

07_boolean_2/main.go
riraum: Done in subsequent commit.

07_boolean_2/main.go
riraum: Should be fixed.

07_boolean_2/main.go
riraum: Makes sense, done.

07_boolean_2/main.go
riraum: Done.

07_boolean_2/main.go
```diff
@@ -15,56 +15,128 @@ E.g.:
...
+	// if bool1 == true && bool2 == true {
+	// 	fmpt.Println("true")
+	// }
+	// Direct print statements
+	// AND // true, true == true || true, false == false || false, false == false || false, true == false
+	fmt.Println("AND", bool1 && bool2)
+	// OR // true, true == true || true, false == true || false, false == false || false, true == true
+	fmt.Println("OR", bool1 || bool2)
+	// NAND // true, true == false || true, false == true, || false, false == true || false, true == true
+	fmt.Println("NAND", !bool1 || !bool2)
```
nobe4, line 63: Adding an alternative
```suggestion
	fmt.Println("NAND", !bool1 || !bool2)
	fmt.Println("NAND", !(bool1 && bool2))
```

07_boolean_2/main.go
riraum: Done

07_boolean_2/main.go
riraum: Done

07_boolean_2/main.go
riraum: Added, thanks!

riraum:
Sorry, still fixing the calcGrade logic. Rest is done I think.

07_boolean_2/main.go
```diff
@@ -15,56 +15,262 @@ E.g.:
...
+//   - If any grade is below 60, return "C-"
+
+// if grade.. > 100 || grade.. < 0 return invalid
+func gradeValid(grade1 int, grade2 int, grade3 int, grade4 int, grade5 int) bool {
+	if (grade1 < 0 || grade1 > 100) || (grade2 < 0 || grade2 > 100) || (grade3 < 0 || grade3 > 100) || (grade4 < 0 || grade4 > 100) || (grade5 < 0 || grade5 > 100) {
+		return false
+	} else {
+		return true
+	}
+}
```
nobe4: This is correct, but not great on the eye.

An alternative is to do "early" returns, e.g. for 3 grades:

```go
func gradeValid(grade1, grade2, grade3) bool {
  if grade1 < 0 || grade1 > 100 {
    return false  
  }

  if grade2 < 0 || grade2 > 100 {
    return false  
  }

  if grade3 < 0 || grade3 > 100 {
    return false  
  }

  return true
}
```

07_boolean_2/main.go
```diff
@@ -15,56 +15,262 @@ E.g.:
...
+
+// check if any grade is < x
+func anyGradeUnder(grade1 int, grade2 int, grade3 int, grade4 int, grade5 int, limit int) bool {
+	if grade1 < limit || grade2 < limit || grade3 < limit || grade4 < limit || grade5 < limit {
+		return true
+	} else {
+		return false
+	}
+}
```
nobe4: Similar idea here, try to do an early return pattern.

07_boolean_2/main.go
```diff
@@ -15,56 +15,262 @@ E.g.:
...
+
+// check if all grades are >= x
+func allGradesOver(grade1 int, grade2 int, grade3 int, grade4 int, grade5 int, limit int) bool {
+	if grade1 >= limit && grade2 >= limit && grade3 >= limit && grade4 >= limit && grade5 >= limit {
+		return true
+	} else {
+		return false
+	}
+}
```
nobe4: Similar idea here, try to do an early return pattern.

07_boolean_2/main.go
```diff
@@ -15,56 +15,262 @@ E.g.:
...
+
+// check if grade avg >= limit
+func gradeAvgOver(grade1 int, grade2 int, grade3 int, grade4 int, grade5 int, limit int) bool {
+	if (grade1+grade2+grade3+grade4+grade5)/5 >= limit {
+		return true
+	} else {
+		return false
+	}
+}
```
nobe4: Here, notice that `(grade1+grade2+grade3+grade4+grade5)/5 >= limit` is a boolean expression that will return either `true` or `false`.

So the function can be rewriten as:

```go
func gradeAvgOver(grade1 int, grade2 int, grade3 int, grade4 int, grade5 int, limit int) bool {
  var avg int = (grade1+grade2+grade3+grade4+grade5)/5 
	return avg >= limit
}
```

I choose to make an extra variable to really make clear what the math is, it's optional and subject to taste.

07_boolean_2/main.go
```diff
@@ -15,56 +15,262 @@ E.g.:
...
+
+func calcGrade(grade1 int, grade2 int, grade3 int, grade4 int, grade5 int) string {
```
nobe4, line 174: In this function, you are doing an antipattern. If you have an expression that gives you a boolean, it's useless to compare it to `true` or `false`. It might even make the code harder to read.


Consider: 
```go
if (x > 2) == true {
  // do something
}
```

vs 

```go
if x > 2 {
  // do something
}
```

The second one is _much_ clearer.

Similarly:

```go
if isGreat() == true {
  // do something
}
```

vs

```go
if isGreat() {
  // do something
}
```

and

```go
if isSmall() == false {
  // do something
}
```

vs

```go
if !isSmall() {
  // do something
}
```

07_boolean_2/main.go
```diff
@@ -15,56 +15,262 @@ E.g.:
...
+		return "invalid grade"
+	}
+	if gradeAvgOver(grade1, grade2, grade3, grade4, grade5, 90) == true {
+		if allGradesOver(grade1, grade2, grade3, grade4, grade5, 80) == true {
+			return "A+"
+		} else if allGradesOver(grade1, grade2, grade3, grade4, grade5, 70) == true {
+			return "A"
+		} else if anyGradeUnder(grade1, grade2, grade3, grade4, grade5, 70) == true {
+			return "A-"
+		}
```
nobe4: Something that you're doing here that can be improved.

Each time you `return`, the execution of the program is stopped and goes back up to the parent function.

So if you `return` from an `if` statement, you don't need `else` or `else if`.

E.g. (don't worry about the logic itself, it's meaningless):

```go
func check1(x int, y int) bool {
	if x == y {
		return false
	} else {
		if x*2 > y {
			return false
		} else if x*3 > y {
			return true
		}
	}
	return false
}

func check2(x int, y int) bool {
	if x == y {
		return false
	}

	if x*2 > y {
		return false
	}
	
	if x*3 > y {
		return true
	}
	
	return false
}
```

See how `check1` has 2 levels of `if`/`else`, whereas the `check2` has only one. The second one reads _much_ better and is strictly equivalent.

Have a look at https://www.youtube.com/watch?v=CFRhGnuXG-4 for inspiration. The video is very technical, focus on the "look and feel" of the code to see the pattern instead of what the code exactly does.

07_boolean_2/main.go
```diff
@@ -15,56 +15,262 @@ E.g.:
...
+	} else if side3 <= side1 && side2 > side3 && side2 > side1 || side1 <= side3 && side2 > side3 && side2 > side1 {
+		if side3+side1 <= side2 {
+			return true
+		}
+	} else if side2 <= side3 && side1 > side2 && side1 > side3 || side3 <= side2 && side1 > side2 && side1 > side3 {
+		if side2+side3 <= side1 {
+			return true
+		}
+	}
```
nobe4: Here, consider creating 2 helper functions:

```go
func sumTwoBiggestSides(a int, b int, c int) int {
  // sums the two biggest numbers between a, b, and c
}

func smallestSide(a int, b int, c int) int {
  // returns the smallest value between a, b, and c
}
```

Your function will then be:

```go
func sumLength(side1 int, side2 int, side3 int) bool {
  var sum int = sumTwoBiggestSides(side1, side2, side3)
  var smallest int = smallestSide(side1, side2, side3)

  return sum >= smallest
}
```

Hopefully you'll see how much more readable this code becomes. The logical operations are put into their own method so that the function can rely on them to abstract the actual logic.

07_boolean_2/main.go
riraum: Thank you! Done. It looks so much better (especially on my smaller screen).

07_boolean_2/main.go
riraum: Done

07_boolean_2/main.go
riraum: Done

07_boolean_2/main.go
riraum: Thank you, done, I see your point!

07_boolean_2/main.go
riraum: I did get a notice `should omit comparison to bool constant, can be simplified to !gradeValid(grade1, grade2, grade3, grade4, grade5) (S1002)go-staticcheck`

I understand your explanation. Thanks and done.
I can just do `if function` if I want it to run on `true` and `if !function` if I want it to run on `false`.

07_boolean_2/main.go
```diff
@@ -15,56 +15,317 @@ E.g.:
...
+	if grade4 < limit {
+		return false
+	}
+
+	if grade5 < limit {
+		return false
+	}
+
+	return true
+}
```
nobe4: Look closely at those two methods, you should see a pattern, and maybe a way to refactor one of the two method.

07_boolean_2/main.go
```diff
@@ -15,56 +15,343 @@ E.g.:
...
+
+	if side1 > side3 && side3 > side2 {
+		return side1 + side3
+	}
+
+	if side2 > side3 && side3 > side1 {
+		return side2 + side3
+	}
+	// Return statement needed, placeholder/error message added
+	return 1337
```
nobe4: this will never be reached, you can remove the last `if` and return directly the last possible sum.

07_boolean_2/main.go
```diff
@@ -15,56 +15,343 @@ E.g.:
...
+
+	if side1 > side3 && side3 > side2 {
+		return side2
+	}
+
+	if side2 > side3 && side3 > side1 {
+		return side1
+	}
+	// Return statement needed, placeholder/error message added
+	return 1337
```
nobe4: this will never be reached, you can remove the last if and return directly the last possible side.

07_boolean_2/main.go
```diff
@@ -15,56 +15,343 @@ E.g.:
...
+	var smallest int = smallestSide(side1, side2, side3)
+	return sum >= smallest
+}
+
+// all sides equal == true
+func allSidesEqual(side1 int, side2 int, side3 int) bool {
+	if side1 == side2 && side3 == side1 {
+		return true
+	}
+	return false
```
nobe4: How can you simplify this (and the next 2 functions)?

07_boolean_2/main.go
riraum: Good one, done

07_boolean_2/main.go
riraum: Removed.

07_boolean_2/main.go
riraum: Removed.

07_boolean_2/main.go
riraum: Why not `sumTwoShortestSides` and `longestSide` according to the requirements?
I did both, but so far it my test for this requirement get me `scalene` instead of `invalid`.

07_boolean_2/main.go
riraum: Thanks, done!

07_boolean_2/main.go
nobe4: Oh yeah, you're right. I got confused!

riraum:
All the checks I wrote seem correct now, finally.

07_boolean_2/main.go
```diff
@@ -15,56 +15,283 @@ E.g.:
...
+		}
+		if allGradesOver(grade1, grade2, grade3, grade4, grade5, 60) {
+			return "C"
+		}
+		if !allGradesOver(grade1, grade2, grade3, grade4, grade5, 60) {
+			return "C-"
+		}
+	}
+	return "grade calc logic not yet implemented for these grades"
+}
```
riraum, line 215: Added this, as I ran into grade combos not being inside my conditions (not being invalid either). Seems needed, according to the specs.

nobe4 (approved):
:tada:

PR #12, opened 2024-10-12
08 arrays

riraum:


08_arrays/main.go
```diff
@@ -43,40 +43,173 @@ func main() {
...
+
+	// for {
+	// 	count4 > 9 {
+	// 		break
+	// 	}
+	// 	fillerMultiplier = intArray[count4] * filler
+	// 	fmt.Println(fillerMultiplier)
+	// 	count4++
+	// }
+
```
riraum: @nobe4 Bit stuck here, any hints?
I'm thinking I might have to loop twice and use append somehow, to get the "x" into the array. But how to put everything together, I don't get right now.

nobe4: Try to divide the task in 2 smaller tasks.

What would be the first and second task?

riraum: Loop through `intArray` to concatenate `x` in a loop as often as the respective element.
Add/append the result to the new array `lengthIntArray`.

nobe4: Yeah, I would have put it slightly differently but you have it essentially:

1. build a string by concatenating `x` multiple times together
2. loop through the array to create a new one with the result of `1.`

`1.` can be made into a function (maybe you've even already done something similar?)

riraum: I think I have both 1. and 2.
I'm now facing the issue again that I can't declare a function inside the another/the main function.
I tried closing the main function and moving the function and the required parts outside with it and then calling it in the loop.
Seems a bit harder compared to `06_functions` where I could just call the function inside the main function.
I do have to move the function outside and close `main` first, no?

nobe4: yes, you can't define a function inside another one. Move it out of the main function.

08_arrays/main.go
```diff
@@ -43,45 +43,213 @@ func main() {
...
+	stringArray[8] = "A"
+	stringArray[9] = "POTATO"
+	fmt.Println(stringArray)
+	fmt.Println("len test=", len(stringArray[0]))
+
+	var lengthStringArray [10]int
+	var count3 = 0
+
+	for {
+		if count3 > 9 {
```
nobe4: Use the length of the array here.

08_arrays/main.go
```diff
@@ -43,45 +43,213 @@ func main() {
...
+
+	// create new empty array of 10 strings
+	var lengthIntArray []string
+	// counter
+	var arrayAssemblyCounter = 0
+
+	// loop to create output array
+	for {
+		// if counter is > than the last/[10] array element break the loop
+		if arrayAssemblyCounter > intArray[arrayAssemblyCounter]-1 {
```
nobe4: This condition can be simplified.

08_arrays/main.go
```diff
@@ -43,45 +43,213 @@ func main() {
...
+// 	return arrayElementOne + arrayElementTwo
+// }
+
+// intArray function below
+// create empty variable for "x" string
+var arrayString string = ""
+
+// counter
+var concatCounter = 0
+
```
nobe4: Those variables need to be in the function.

08_arrays/main.go
```diff
@@ -43,45 +43,231 @@ func main() {
...
+	var count int = 1
+	var times int = 10
+
+	for {
+		if count > times {
+			break
+		}
+		tenInt = append(tenInt, count)
+		count++
+	}
```
nobe4: Here you can use a single variable for the counter.

08_arrays/main.go
```diff
@@ -43,45 +43,231 @@ func main() {
...
+	fmt.Println(tenInt[len(tenInt)-1])
 
 	// TODO: Using the array from the previous step, create a new array
 	// containing the first 5 elements multiplied by 2.
 	// Hint: Use a loop.
 	// Hint: The new array should contain [2, 4, 6, 8, 10].
 	// Print the new array.
 
+	// create copy of array
+	// loop 5 times (first 4 elements needed)
```
nobe4:
```suggestion
	// loop 5 times (first 5 elements needed)
```

08_arrays/main.go
riraum: It did seem to run like this, but moved them!

08_arrays/main.go
```diff
@@ -43,45 +43,230 @@ func main() {
...
+	var multCounter = 0
+	for {
+		// TODO: replace hard coded value with last index of
+		if multCounter > 9 {
+			break
+		}
+		multArray = append(multArray, anotherIntArray[multCounter]*yetAnotherIntArray[multCounter])
+		// increase counter
+		multCounter++
+	}
```
nobe4: make this with just one loop, the extra 2 are redundant

riraum: Done, good one. Felt a bit dirty writing it, thought of variables, but this is better.

08_arrays/main.go
riraum: Done. Mixed it up with the element position of the array.

08_arrays/main.go
riraum: Done.

08_arrays/main.go
riraum: Yeah that bugged me, had more like this and was on my TODO list. Done.

08_arrays/main.go
riraum: Of course, done.

nobe4 (approved):

PR #13, opened 2024-10-16
Print length of map human

riraum:
09_maps exercise.

### DIFF
--- a/09_maps/main.go
+++ b/09_maps/main.go
@@ -135,19 +135,19 @@ func main() {
 	// print people[0]
 	fmt.Println(people)
 	// print BMI
-	calcBMI(people[0])
-	// var counter int = 0
+	// calcBMI(people[0])
+	var counter int = 0
 	// // var calcPeople []map[string]int = people[counter]
 
-	// for {
-	// 	if counter >= len(people) {
-	// 		break
-	// 	}
-	// 	// print bmi by calling the calcBMI with an array value of people
-	// 	// calcBMI(calcPeople)
-	// 	calcBMI(people[counter])
-	// 	counter++
-	// }
+	for {
+		if counter >= len(people) {
+			break
+		}
+		// print bmi by calling the calcBMI with an array value of people
+		// calcBMI(calcPeople)
+		calcBMI(people[counter])
+		counter++
+	}
 
 	// For later exercises, use the same format as the previous one.
 
@@ -196,14 +196,17 @@ func createNewMap(mapInput map[string]int) {
 func calcBMI(mapInput map[string]int) {
 	var value int
 	var exists bool
-	var bmi float32
-	var nonFloatBmi float64
+	var bmi float64
+	// var nonFloatBmi float64
 	var heightAdjust float64
 
 	var heightConvert int = mapInput["height"]
 
-	var heighConverted float64 = float64(heightConvert)
+	var heightConverted float64 = float64(heightConvert)
 
+	var weightConvert int = mapInput["weight"]
+
+	var weightConverted float64 = float64(weightConvert)
 	value, exists = mapInput["height"]
 	if !exists || value == 0 {
 		fmt.Println("Height not found")
@@ -218,19 +221,20 @@ func calcBMI(mapInput map[string]int) {
 	// bmi = mapInput["weight"] / (mapInput["height"]/100 * mapInput["height"]/100)
 	// bmi = mapInput["weight"] / (heighAdjust * heighAdjust)
 	// heightConvert = float64((mapInput["height"])
-	heightAdjust = heighConverted / 100
+	heightAdjust = heightConverted / 100
+	// debug
+	fmt.Println(heightAdjust)
+	fmt.Println(weightConverted)
 	// bmi = (mapInput["weight"]) / heighAdjust * heighAdjust
+	// bmi = (weightConverted) / (heightAdjust) * int(heightAdjust)
 	// bmi = 70 / ((180 / 100) * (180 / 100))
 	// bmi = (70 / (1.80 * 1.80))
-	nonFloatBmi = (70 / (180 * 180))
+	bmi = (weightConverted / (heightConverted * heightConverted))
 	// debug
-	fmt.Println("nonfloatBMI is", nonFloatBmi)
-	nonFloatBmi = nonFloatBmi * 10000
-	fmt.Println("nonfloatBMI is", nonFloatBmi)
-
+	bmi = bmi * 10000
 	// debug
 	// fmt.Println(bmi)
-	fmt.Println("nonfloatBMI is", nonFloatBmi)
+	// fmt.Println("nonfloatBMI is", nonFloatBmi)
 	// TODO: check and fix syntax for name and bmi variable
 	fmt.Println("BMI is", bmi)
 }
--- a/09_maps/main.go
+++ b/09_maps/main.go
@@ -176,7 +176,7 @@ func main() {
 	fmt.Println(largestInt([]int{0, 3000, 0, -100})) // 3000
 
 	// test helper string function
-	fmt.Println(largestString([]string{"", ""}))                         // ""
+	// fmt.Println(largestString([]string{"", ""}))                         // ""
 	fmt.Println(largestString([]string{"a", "bc", "def"}))               // "def"
 	fmt.Println(largestString([]string{"qwertyuiop", "xyz", "xyzabcd"})) // "qwertyuiop"
 	fmt.Println(largestString([]string{"a", "bcd", "ef"}))               // "bcd"
@@ -335,38 +335,88 @@ func largestInt(s []int) int {
 // write a function that takes a list of string and return the longest string, or an empty string.
 func largestString(s []string) string {
 	// counter
-	var counter int
+	// var counter int
 	// length of slice counter
 	// var lengthSlice int = len(s)
 	// initialize largest
-	var largestStringResult string
-	// get length of s[0], assign to variable
-	// var largestInitial int = len(s[0])
+	// var largestStringResult string
+	// get length of slice
+	// lenSlice := len(s)
 	// if array of strings is empty, return empty string
 	if len(s) == 0 {
 		return ""
 	}
 	// loop
-	for {
-		// if counter is higher than the length of array, break
-		if counter > len(s)-1 {
-			break
-		}
-		// if length of element of array is higher than 0
-		if len(s[counter]) > len(s[0]) {
-			// assign string to variable for output
-			largestStringResult = s[counter]
-		}
-		// if len(s[counter2]) > 0 {
-		// 	// assign string to variable for output
-		// 	largestStringResult = s[counter2]
-		// }
-		// increase counter
-		counter++
+	// for {
+	// 	// if counter is higher than the length of array, break
+	// 	if counter > len(s)-1 {
+	// 		break
+	// 	}
+	// // 	if len(s[counter]) >
+	// 	// if length of element of array is higher than 0
+	// 	// if len(s[counter]) > 0 {
+	// 	// 	// assign string to variable for output
+	// 	// 	largestStringResult = s[counter]
+	// 	// }
+	// 	// if len(s[counter2]) > 0 {
+	// 	// 	// assign string to variable for output
+	// 	// 	largestStringResult = s[counter2]
+	// 	// }
+	// 	// increase counter
+	// 	counter++
+	// }
+
+	// hard coded comparison of slice elements
+	if len(s[0]) > len(s[1]) && len(s[0]) > len(s[2]) {
+		return s[0]
 	}
+	if len(s[1]) > len(s[0]) && len(s[1]) > len(s[2]) {
+		return s[1]
+	}
+	if len(s[2]) > len(s[0]) && len(s[2]) > len(s[1]) {
+		return s[2]
+	}
+	return "error"
+
 	// return final, largest result
-	return largestStringResult
+	// return largestStringResult
 }
+
+// helper loop
+// get length of slice
+// compare
+// func helperLoop(s []string) string {
+// 	// 	var counter int
+// 	// for {
+// 	// 	if counter > len(s) {
+// 	// 		break
+// 	// 	}
+// 	if len(s[0]) > len(s[1]) && len(s[0]) > len(s[2]) {
+// 		return s[0]
+// 	}
+// 	if len(s[1]) > len(s[0]) && len(s[1]) > len(s[2]) {
+// 		return s[1]
+// 	}
+// 	return s[2]
+// }
+
+// func largestString(s []string) string {
+// 	// 	var counter int
+// 	// for {
+// 	// 	if counter > len(s) {
+// 	// 		break
+// 	// 	}
+// 	if len(s[0]) > len(s[1]) && len(s[0]) > len(s[2]) {
+// 		return s[0]
+// 	}
+// 	if len(s[1]) > len(s[0]) && len(s[1]) > len(s[2]) {
+// 		return s[1]
+// 	}
+// 	if len(s[2]) > len(s[0]) && len(s[2]) > len(s[1]) {
+// 		return s[2]
+// 	}
+// 	return "error"
+// }
 
 // main function
 func highestBMI(arrayOfMapInput []map[string]int) map[string]int {
--- a/09_maps/main.go
+++ b/09_maps/main.go
@@ -150,6 +150,9 @@ func main() {
 	// calculate average BMI
 	fmt.Println("Avg BMI is", calcAvgBMI(people))
 
+	// calculate highest BMI
+	fmt.Println("Highest BMI is", highestBMI(people))
+
 	// Close main function, to be able to declare another function
 }
 
@@ -272,6 +275,31 @@ func calcAvgBMI(arrayOfMapInput []map[string]int) float64 {
 
 // TODO: Create a function that takes an array of maps as an argument and
 // returns the map with the highest BMI.
+
+// create function
+// accept array of maps as input
+// calculate the BMI with function calcBMI for all elements of the array, in a loop
+// compare all elements with each other to get highest value/BMI
+// return map of highest value
+
+func highestBMI(arrayOfMapInput []map[string]int) map[string]int {
+	var counter1 int = 0
+	var counter2 int = 1
+	// for {
+	// var result1 = arrayOfMapInput[counter1]
+	// var result2 = arrayOfMapInput[counter2]
+	// if counter1 >= len(arrayOfMapInput) {
+	// 	break
+	// }
+	if calcBMI(arrayOfMapInput[counter1]) > calcBMI(arrayOfMapInput[counter2]) {
+		return arrayOfMapInput[counter1]
+	}
+	return arrayOfMapInput[counter2]
+
+	// counter1++
+}
+
+// if calcBMI()
 
 // TODO: Create a function that takes an array of maps as an argument and
 // add as a new key-value pair to each map the key "bmi" with the value
--- a/09_maps/main.go
+++ b/09_maps/main.go
@@ -92,14 +92,11 @@ func main() {
 
 	// TODO: Create a new map called "person" with the following key-value
 	// pairs:
-	// - "name" -> "John"
 	// - "age" -> 25
 	// - "height" -> 180
 	// - "weight" -> 70
 	// Print the map and its length.
 	var person map[string]string = map[string]string{}
-	person["name"] = "John"
-	// TODO: add rest of the values as string/int pair instead of string/string
 	person["age"] = "25"
 	person["height"] = "180"
 	person["weight"] = "70"
--- a/09_maps/main.go
+++ b/09_maps/main.go
@@ -128,19 +128,24 @@ func main() {
 	// people[0]["height"] = 180
 	// people[0]["weight"] = 70
 	people = append(people, map[string]int{"height": 180, "weight": 70})
+	people = append(people, map[string]int{"height": 160, "weight": 50})
+	people = append(people, map[string]int{"height": 170, "weight": 80})
+	people = append(people, map[string]int{"height": 150, "weight": 60})
+	people = append(people, map[string]int{"height": 190, "weight": 90})
 	// print people[0]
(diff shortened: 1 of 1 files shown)


09_maps/main.go
```diff
@@ -85,6 +97,14 @@ func main() {
...
 	// - "weight" -> 70
 	// Print the map and its length.
+	var person map[string]string = map[string]string{}
+	person["name"] = "John"
+	// TODO: add rest of the values as string/int pair instead of string/string
+	person["age"] = "25"
+	person["height"] = "180"
+	person["weight"] = "70"
+	fmt.Println(person)
+	fmt.Println(len(person))
```
riraum: @nobe4 Should I use `interface{}` in order to be able to create a map with mixed key value pairs?

nobe4: In theory yes, but let's not talk about `interface{}` for now.
Use strings everywhere for this exercise :)

nobe4: Actually, I'll need to update my exercises, they have some mixed values.
Sorry about that

riraum: Okay, sounds good.
I was struggling a bit how to proceed due to that. Especially later with the multiplication...

nobe4: I've pushed changes to the master branch, please merge it into your branch to have (hopefully) less confusing exercises.

09_maps/main.go
```diff
@@ -94,35 +111,76 @@ func main() {
...
+		newMap["age"] = mapInput["age"] + 2
+		newMap["height"] = mapInput["height"] * 2
+		newMap["weight"] = mapInput["weight"] * 2
+		if mapInput["height"] > 200 {
+			newMap["tall"] = 1
+		}
+		newMap["tall"] = 0
+	}
+	fmt.Println(mapInput)
+}
```
riraum, line 153: @nobe4 Does this task also need adjustment for mix of types, or do I have to find a way?

09_maps/main.go
```diff
@@ -94,35 +111,76 @@ func main() {
 	// - "tall" -> 1 if height > 200, 0 otherwise
```
riraum, line 116: @nobe4 Does this task also need adjustment for mix of types, or do I have to find a way?

nobe4: `1` and `0` are numbers, so you can use `int` as for the other values.

09_maps/main.go
```diff
@@ -155,27 +155,22 @@ func createNewMap(mapInput map[string]int) {
...
-		if !exists {
-			fmt.Println("Height not found")
-		}
-		key, exists = mapInput["weight"]
-		if !exists {
-			fmt.Println("Weight not found")
-		}
-		bmi = mapInput["weight"] / (mapInput["height"] * mapInput["height"])
+	value, exists = mapInput["height"]
+	if !exists && value == 0 {
```
nobe4: `!exists` implies `value == 0`, so  using `&&` here is redundant.
However, using writing `!exist || value == 0`, would make more sense.

riraum: I tried to only do it with a boolean check, but that wasn't enough to run. 
Changed to your suggestion.

09_maps/main.go
```diff
@@ -111,18 +122,99 @@ func main() {
...
+	// var calcPeople = people[counter]
+
+	// for {
+
+	// 	if counter > len(people) {
+	// 		break
+	// 	}
+	// 	fmt.Println(calcBMI(calcPeople))
+	// 	counter++
+	// }
```
riraum: @nobe4 The syntax to fill the array doesn't seem correct, I get `panic: runtime error: index out of range [0] with length 0`.
Should I look into using `append` here?

nobe4: Yes

09_maps/main.go
```diff
@@ -111,18 +122,129 @@ func main() {
...
+	// debug
+	fmt.Println("nonfloatBMI is", nonFloatBmi)
+	nonFloatBmi = nonFloatBmi * 10000
+	fmt.Println("nonfloatBMI is", nonFloatBmi)
+
+	// debug
+	// fmt.Println(bmi)
+	fmt.Println("nonfloatBMI is", nonFloatBmi)
+	// TODO: check and fix syntax for name and bmi variable
+	fmt.Println("BMI is", bmi)
```
riraum: @nobe4 Tried to get a result by using float variables and multiplying the result by `*10000`.
For some reason it only seems to work with the height in float numbers.

Also tried to convert the height the height converted to a float variable to be able to then divide by 100.
I get`declared and not used: heightAdjustcompiler` even though I'm trying to use it in line 221.
How should I proceed?

nobe4: - using `float64` for all numerical values in this exercise is a good idea, you can change the map and function accordingly. But you can also just convert the `int` to `float64` like you've done here.
- `Also tried to convert the height the height converted to a float variable to be able to then divide by 100.` => I have no idea what you're trying to say haha
- `declared and not used: heightAdjust` is showing up because you're only ever setting the value of this variable, never using it.

09_maps/main.go
```diff
@@ -111,18 +122,129 @@ func main() {
...
+// BMI = weight / (height * height)
+// If the map does not have the key "height", print "Height not found".
+// If the map does not have the key "weight", print "Weight not found".
+// Otherwise, print "<name>'s BMI is <bmi>".
+// Hint: Use this function in later exercises.
+
+func calcBMI(mapInput map[string]int) {
+	var value int
+	var exists bool
+	var bmi float32
```
nobe4: Why not `float64`?

riraum: Didn't read too much into the difference of `float32` vs `float64`.
Changed.

09_maps/main.go
```diff
@@ -223,6 +226,31 @@ func calcBMI(mapInput map[string]int) {
...
+	for {
+		if counter > len(arrayOfMapInput) {
+			break
+		}
+		bmiSum += calcBMI(arrayOfMapInput[counter])
+		counter++
+	}
+	avgBMI = bmiSum / lenConverted
+	fmt.Println(avgBMI)
+}
```
riraum: @nobe4 Unsure how to correctly run the `calcBMI` function inside the `calcAvgBMI` one.
Currently I get `calcBMI(arrayOfMapInput[counter]) (no value) used as valuecompiler[TooManyValues])` as error on line 247.
It seems I'm putting too many values into the function call.

Do I maybe have to remove the loop out of the function and separate the function call?
I did attempt to put the `calcBMI` function call into a variable within the function, but that doesn't seem to solve it.

nobe4: Your function `calcBMI` doesn't return a value, so you should change that first.

09_maps/main.go
```diff
@@ -221,6 +223,16 @@ func calcBMI(mapInput map[string]int) {
...
+
+	for {
+		var counter int = 0
+		if counter > len(mapInput) {
+			break
+		}
+		counter++
+		return bmi
+	}
+
```
riraum: @nobe4 I tried that, gives me `too many return values have (float64) want ()` same if I do `return bmi`.
I tried to just have `return`, that gave me `redundant return statement (S1023)` no matter if with or no loop.

nobe4: Go back to the function exercise and re-read how functions have to be defined.

09_maps/main.go
riraum, line 1: @nobe4 My understanding is that I somehow have to compare one element with ALL other elements and only if that element is bigger than ALL others it should get returned.
I did start an attempt at a loop, thought of maybe creating a helper function, tried to visualize it with notes and finally with this hard coded element comparison, which obviously is not the goal.
Somehow combining loops sounds good to me, but to piece all of that together is a bit hard right now.
Am I way off or in the right direction?

09_maps/main.go
```diff
@@ -111,18 +122,194 @@ func main() {
...
+	// 	}
+	// 	if calcBMI(arrayOfMapInput[counter1]) > calcBMI(arrayOfMapInput[counter2]) {
+	// 		counter1++
+	// 	}
+	// 	// if calcBMI(arrayOfMapInput[counter1])
+	// }
+	// test return
+	return arrayOfMapInput[0]
+	// return arrayOfMapInput[counter1]
+}
```
riraum: @nobe4 My understanding is that I somehow have to compare one element with ALL other elements and only if that element is bigger than ALL others it should get returned.
I did start an attempt at a loop, thought of maybe creating a helper function, tried to visualize it with notes and finally with this hard coded element comparison, which obviously is not the goal.
Somehow combining loops sounds good to me, but to piece all of that together is a bit hard right now.
Am I way off or in the right direction?

nobe4: Try the following:

- write a function that takes a list of int and return the biggest, or 0.

  ```go
  func largestInt(s []int) int {
    // ???
  }
  fmt.Println(largestInt([]int{}) // 0
  fmt.Println(largestInt([]int{1,2,3}) // 3
  fmt.Println(largestInt([]int{-1,-2,0}) // 0
  fmt.Println(largestInt([]int{100,0,3000,0,-100}) // 3000
  ```

- write a function that takes a list of string and return the longest string, or an empty string.

  ```go
  func largestString(s []string) string {
    // ???
  }
  fmt.Println(largestInt([]string{"", ""}) // ""
  fmt.Println(largestInt([]string{"a", "bc", "def"}) // "def"
  ```

Both functions should have the same pattern, which I can give you if you are too stuck.
Use the same pattern to find the map with the largest BMI.

riraum: Thank you.
I did find some loop logic example that I tried to change to the loop syntax I previously used.
The `largestInt` does return `0`, then it gets stuck. Might have to give it another try later.
I did try to use the logic in `highestBMI`, also not successfully yet, but maybe closer than before.
Not sure if more hints or just a break is best at this point.

09_maps/main.go
```diff
@@ -111,18 +122,261 @@ func main() {
...
+	if len(s) == 0 {
+		return ""
+	}
+	var largest = s[0]
+	for {
+		var counter int
+		if counter > len(s) {
+			break
+		}
+		if s[counter] > largest {
```
nobe4: This is incorrect, you need to compare string length

09_maps/main.go
nobe4: I added a comment, but maybe you need to also split the problem into smaller problems?

09_maps/main.go
```diff
@@ -69,22 +69,39 @@ func main() {
...
 	// Do the same for the key "weight"
+	var value int
+	var exists bool
+	value, exists = human["age"]
+	if exists {
+		fmt.Println("age exists in the map")
+	}
+	if value == 0 {
+		fmt.Println("age does not exist in the map")
+	}
```
nobe4: This is incorrect, `if value == 0` means that the value is equal to `0`, not that it doesn't exist in the map. Using the variable `exists` that you defined above is the only way to check if the value exists.

09_maps/main.go
```diff
@@ -69,22 +69,39 @@ func main() {
 	fmt.Println(human)
 
 	// TODO: Print the length of the map human
+	fmt.Println(len(human))
 
 	// TODO: Check if the key "age" exists in the map human
 	// If so print "age exists in the map"
 	// If not print "age does not exist in the map"
 	// Do the same for the key "weight"
```
nobe4, line 77: You missed this one

09_maps/main.go
riraum: Should be fixed.

09_maps/main.go
```diff
@@ -111,18 +122,268 @@ func main() {
...
+// create function
+// create counter
+// create loop
+// create break condition if counter is > than the key-value pairs the map has
+// operate on the input value of the keys and do math on it
+// add the result to a new map
+// add a new key value element and fill it depending on the value of the input height key
+// print new map
+
+var newMap map[string]int = map[string]int{}
```
nobe4: This should be defined within the function that uses it.
I'll get an exercises about `scope` done later, it seems to be a recurring issue :)

09_maps/main.go
```diff
@@ -111,18 +122,268 @@ func main() {
...
+		if counter > len(mapInput) {
+			break
+		}
+		newMap["age"] = mapInput["age"] + 2
+		newMap["height"] = mapInput["height"] * 2
+		newMap["weight"] = mapInput["weight"] * 2
+		if mapInput["height"] > 200 {
+			newMap["tall"] = 1
+		}
+		newMap["tall"] = 0
```
nobe4: This is incorrect.

Print the result of `createNewMap` with a couple of different maps, esp. with one that _should_ be tall, and one that _shouldn't_

09_maps/main.go
```diff
@@ -111,18 +122,268 @@ func main() {
...
+	// fmt.Println("height debug", mapInput["height"])
+	// debug
+	// fmt.Println(heightAdjust)
+	// fmt.Println(weightConverted)
+	bmi = (weightConverted / (heightConverted * heightConverted))
+	bmi = bmi * 10000
+	// debug
+	// fmt.Println(bmi)
+	// fmt.Println("BMI is", bmi)
+	return bmi
```
nobe4: Couple of comments
- Don't use `value`, but rather the variable containing the corresponding value (here `heightConvert`/`weightConvert`
- `heightConvert`/`weightConvert` are weird names, if you want to differentiate from `int` and `float`, using `heigthInt` and `heigthFloat` might be slightly better.
- Convert those to `float64` only _after_ you are sure they both exist

09_maps/main.go
```diff
@@ -111,18 +122,268 @@ func main() {
...
+// add the sum of all array elements
+// increase counter
+// divide through the amount of array elements
+// print result = avg BMI
+func calcAvgBMI(arrayOfMapInput []map[string]int) float64 {
+	var counter int
+	var bmiSum float64
+	var avgBMI float64
+	var lenConvert int = len(arrayOfMapInput)
+	var lenConverted float64 = float64(lenConvert)
```
nobe4: Name and simplification suggestion
```suggestion
	var mapLen float64 = float64(len(arrayOfMapInput))
```

09_maps/main.go
```diff
@@ -111,18 +122,268 @@ func main() {
...
+	// loop
+	for {
+		// if counter is higher than the length of array, break
+		if counter > len(s)-1 {
+			break
+		}
+		// if length of element of array is higher than variable largest, re assign that length of value to largest
+		if len(s[counter]) > largestInitial {
+			largestStringResult = s[counter]
+		}
```
nobe4: This is not correct, this compares the length of the current string with the length of the first string in the list.

E.g.
```go
largestString([]string{"a", "abcd", "ab"}) // returns "ab", but should return "abcd"
```

riraum: Should be fixed.

09_maps/main.go
riraum: Done!

09_maps/main.go
riraum: Makes sense. I think I ran into some issues due to declaring variables before the assigned things existed.

09_maps/main.go
riraum: Should work now.

09_maps/main.go
riraum: Moved!

09_maps/main.go
riraum: What do you mean? It seems to print, `2`, isn't that the maps length?

09_maps/main.go
riraum: If/if (or if/else) statement doesn't run, currently get`declared and not used: value`, but seem to need 2 variables to be able to check correctly.
 
`To check if a key exists in a map, you can use the following syntax:`
seems to be missing in the instructions.
The next mentioned syntax to get the value seemed to work, except it doesn't :(

09_maps/main.go
nobe4: I meant this `// Do the same for the key "weight"`

09_maps/main.go
nobe4: > currently get `declared and not used: value`

This is unrelated to the condition. It just means that you're not using the variable `value`. If you print it, the error will go away.


> seems to be missing in the instructions.

Oh yeah, it's badly written. It should have been:

```
To check if a key exists in a map and to get the value, you can use the following syntax:

    var value int
    var exists bool
    value, exists = m["age"]

```

09_maps/main.go
```diff
@@ -111,18 +134,268 @@ func main() {
...
+		newMap["age"] = mapInput["age"] + 2
+		newMap["height"] = mapInput["height"] * 2
+		newMap["weight"] = mapInput["weight"] * 2
+		if mapInput["height"] > 200 {
+			newMap["tall"] = 1
+		}
+		if mapInput["height"] <= 200 {
+			newMap["tall"] = 0
+		}
```
nobe4: How can you simplify this?

09_maps/main.go
```diff
@@ -111,18 +134,268 @@ func main() {
...
+	value, exists = mapInput["height"]
+	if !exists || value == 0 {
+		fmt.Println("Height not found")
+	}
+	var heightFloat float64 = float64(heightInt)
+	value, exists = mapInput["weight"]
+	if !exists || value == 0 {
+		fmt.Println("Weight not found")
+	}
+	var weightFloat float64 = float64(weightInt)
```
nobe4: You are getting `mapInput["height"]` and `mapInput["weight"]` twice, this is unnecessary.

Can you find a way to simplify this further?

09_maps/main.go
riraum: I thought of using and if, else statement before, but I wanted to try to become a nevernester.
But I guess just one time nesting is okay.

09_maps/main.go
riraum: Done.

09_maps/main.go
riraum: Printed it. What if I don't want to have it printed?

09_maps/main.go
riraum: I commented out both variables, as they are just used to immediately be use for the float64 variables. Not sure you meant that?

09_maps/main.go
nobe4: You would have the same level of nesting using an if/else or two ifs in this case.

Plus, `(x > y)` is strictly equivalent to `!(x <= y)`, so you can simplify:
```
if x > y { // do something }
if x <= y { // do something else }
```
by
```
if x > y { // do something }
else { // do something else }
```

09_maps/main.go
nobe4: There's a way to discard values in go, if you want to read about it: https://go.dev/doc/effective_go#blank

09_maps/main.go
```diff
@@ -111,18 +136,268 @@ func main() {
...
+	if !exists || value == 0 {
+		fmt.Println("Height not found")
+	}
+	var heightFloat float64 = float64(mapInput["height"])
+
+	value, exists = mapInput["weight"]
+	if !exists || value == 0 {
+		fmt.Println("Weight not found")
+	}
+	var weightFloat float64 = float64(mapInput["weight"])
```
nobe4: You are writing `mapInput["height"]` twice, which is one too many.
Same for `mapInput["weight"]`

09_maps/main.go
nobe4: Nop, you are still comparing against the first string in the list.

riraum: Confused, thought I did the change. But now 🥴

09_maps/main.go
riraum: Have done the if/else.

09_maps/main.go
riraum: Have removed the float64 conversion variables. Conversion is now done directly during the re-assignment of `bmi`.

I could also do the float64 conversion variable before I check if `height, weight` exists, then use the variable for the check, but I think that is not preferred before I don't know they exist?

nobe4: Still no, you are still writing `mapInput["weight"]` one too many time.

What you are doing is:
```go
value, exists = mapInput["weight"]
// ...
var weightFloat float64 = float64(mapInput["weight"])
```

What you should do is:
```go
value, exists = mapInput["weight"]
// ...
var weightFloat float64 = float64(value)
```

riraum: I made use of the variable `value` for the `weight/heightFloat` variable now.

09_maps/main.go
nobe4: The code still isn't correct :grimacing:

09_maps/main.go
```diff
@@ -111,18 +136,271 @@ func main() {
...
+	// test helper int function
+	fmt.Println(largestInt([]int{}))                      // 0
+	fmt.Println(largestInt([]int{1, 2, 3}))               // 3
+	fmt.Println(largestInt([]int{-1, -2, 0}))             // 0
+	fmt.Println(largestInt([]int{100, 0, 3000, 0, -100})) // 3000
+
+	// test helper string function
+	fmt.Println(largestString([]string{"", ""}))                         // ""
+	fmt.Println(largestString([]string{"a", "bc", "def"}))               // "def"
+	fmt.Println(largestString([]string{"xyz", "xyzabcd", "qwertyuiop"})) // "qwertyuiop"
```
nobe4: your function `largestString` is still incorrect, here's a few more examples 

```suggestion
	fmt.Println(largestString([]string{"xyz", "xyzabcd", "qwertyuiop"})) // "qwertyuiop"
		fmt.Println(largestString([]string{"a", "bcd", "ef"})) // "bcd"
		fmt.Println(largestString([]string{"a", "b", "c"})) // "a"
		fmt.Println(largestString([]string{"abc", "de", "f"})) // "abc"
		fmt.Println(largestString([]string{"abc", "", "def"})) // "abc"
	  fmt.Println(largestString([]string{"", "abc", ""})) // "abc"
```

riraum: Added

09_maps/main.go
riraum: Now?

nobe4 (approved):
You should do some cleanup, but the code looks correct now 🎉

PR #14, opened 2024-10-24
10_simplification exercise

riraum:
Exercise 10_simplification

### DIFF
--- a/10_simplification/main.go
+++ b/10_simplification/main.go
@@ -207,16 +207,52 @@ func main() {
 	// TODO: write a loop that sums numbers until it reaches 20, print each new
 	// sum
 	// e.g. 0, 0+1, 0+1+2, 0+1+2+3, ...
-
-	// TODO: write a function that sums 3 `int` and returns the result
-	// TODO: write a function that multiplies 3 `int` and returns the result
-	// TODO: write a function that returns the sum of all numbers between its
-	// parameters.
-	// If the second parameter is less than the first, return 0.
-	// e.g. sum(1, 3) => 1 + 2 + 3 = 6
-	//      sum(2, 5) => 2 + 3 + 4 + 5 = 14
+	var sumTwenty int
+	for r := 0; r < 20; r++ {
+		sumTwenty += r
+		fmt.Println(sumTwenty)
+	}
 
 	// TODO: declare an array of 5 integers and print it
 	// TODO: declare an array of 5 strings and print it
 	// TODO: declare a map with 3 key-value pairs and print it
+
+	// close main function
+	fmt.Println("Print sumThreeInt", sumThreeInt(1, 2, 3))
+
+	fmt.Println("Print mulThreeInt", mulThreeInt(1, 2, 3))
+
+	fmt.Println("Print returnBetween", returnBetween(1, 2))
+	fmt.Println("Print returnBetween", returnBetween(1, 9))
+	fmt.Println("Print returnBetween", returnBetween(-1, 2))
+	fmt.Println("Print returnBetween", returnBetween(0, 0))
+	fmt.Println("Print returnBetween", returnBetween(3, 5))
+
+}
+
+// TODO: write a function that sums 3 `int` and returns the result
+func sumThreeInt(a, b, c int) int {
+	return a + b + c
+}
+
+// TODO: write a function that multiplies 3 `int` and returns the result
+func mulThreeInt(a, b, c int) int {
+	return a * b * c
+}
+
+// TODO: write a function that returns the sum of all numbers between its
+// parameters.
+// If the second parameter is less than the first, return 0.
+// e.g. sum(1, 3) => 1 + 2 + 3 = 6
+//
+//	sum(2, 5) => 2 + 3 + 4 + 5 = 14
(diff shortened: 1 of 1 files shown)


10_simplification/main.go
```diff
@@ -145,31 +147,122 @@ can be taken.
...
+		if k%2 == 0 {
+			fmt.Println(k)
+		}
+	}
 	// TODO: write a loop that prints the first 10 odd numbers
+	for l := 0; l < 10; l++ {
+		if l%2 != 0 {
+			fmt.Println(l)
+		}
+	}
```
nobe4: This prints the first odd numbers under 10, the task is to print the first 10 odd numbers.

I.e., you print 1, 3, 5, 7, 9
You need to print 1, 3, 7, 9, 11, 13, 15, 17, 19, 21

10_simplification/main.go
```diff
@@ -145,31 +147,122 @@ can be taken.
...
+		sumFirstTen += m
+	}
+	// debug
+	fmt.Println(sumFirstTen)
 	// TODO: write a loop that calculates the sum of the first 10 odd numbers
+	var sumFirstTenOdd int
+	for n := 0; n < 10; n++ {
+		if n%2 != 0 {
+			sumFirstTenOdd += n
+		}
```
nobe4: Same comment as https://github.com/riraum/forward/pull/14/files#r1814380788

10_simplification/main.go
```diff
@@ -145,31 +147,122 @@ can be taken.
...
+// parameters.
+// If the second parameter is less than the first, return 0.
+// e.g. sum(1, 3) => 1 + 2 + 3 = 6
+//
+//	sum(2, 5) => 2 + 3 + 4 + 5 = 14
+func returnBetween(a, b float64) float64 {
+	var sum float64
+	if b < a {
+		return 0
+	}
```
nobe4: You only need `sum` if you use it, so you can define it later

```suggestion
	if b < a {
		return 0
	}
	
	var sum float64
```

10_simplification/main.go
riraum: Done!

10_simplification/main.go
riraum: Done!

10_simplification/main.go
riraum: Done!

10_simplification/main.go
```diff
@@ -145,31 +147,129 @@ can be taken.
...
+		}
+	}
 	// TODO: write a loop that prints the first 10 odd numbers
+	x := 0
+	for l := 0; x <= 10; l++ {
+		if l%2 == 0 {
+			continue
+		}
+		fmt.Println("Print odd numbers", l)
+		x++
```
nobe4: This is too complex, you don't need `x` and `l`, you can do with just one variable.

10_simplification/main.go
```diff
@@ -145,31 +147,129 @@ can be taken.
...
 	// TODO: write a loop that calculates the sum of the first 10 odd numbers
+	var sumFirstTenOdd int
+	y := 0
+	for n := 0; y < 10; n++ {
+		if n%2 == 0 {
+			continue
+		}
+		sumFirstTenOdd += n
+		y++
+	}
```
nobe4: Same comment here.

10_simplification/main.go
riraum: Took me quite a bit of playing around. Did try various counter placements, no dice.
Had to search the web and after no result seemed to help, read the Brave Search AI answer, that helped.
I guess I do have to create the logic around that after an odd number is followed by an odd number and therefore both increase my break limit and counter to double the values.
Screenshot 2024-10-24 at 20 05 48
I try to stay away from AI for now, could disable the auto results if you think that's better.

10_simplification/main.go
riraum: Same as above, should be done.

10_simplification/main.go
```diff
@@ -171,14 +171,13 @@ func main() {
 		}
 	}
 	// TODO: write a loop that prints the first 10 odd numbers
-	x := 0
-	for l := 0; x <= 10; l++ {
-		if l%2 == 0 {
-			continue
+	for l := 0; l <= 20; l++ {
```
nobe4, line 174: That's not what I had in mind.

Look again at how you solved it in 05_loop.

10_simplification/main.go
nobe4: That's not what I had in mind.

Look again at how you solved it in 05_loop.

riraum: My implementation differed a bit from the Brave Search result.
I think I had looked at 05_loops and did check the `Write a loop that prints the first 100 odd numbers.` again, but there I use 2 variables.
Added that logic too, but with 2 variables.

I did try to remove the continue condition with from this shorter loop syntax, but seems it has to stay there, seem to only be able to add a break statement after.

nobe4: Nice, it's good that you did a couple of additional examples!

nobe4 (approved):

PR #15, opened 2024-10-25
11 structures exercise

riraum:


11_structures/main.go
```diff
@@ -69,41 +73,327 @@ instances inside the function.
...
+
+func longestSide(side1, side2, side3 float64) float64 {
+	if side1 < side2 && side2 < side3 {
+		return side3
+	}
+	if side1 < side3 && side3 < side2 {
+		return side2
+	}
+	return side1
+}
```
nobe4: make this a function of the structure `triangle`

11_structures/main.go
```diff
@@ -69,41 +73,327 @@ instances inside the function.
...
+
+func sumTwoShortestSides(side1, side2, side3 float64) float64 {
+	if side1 < side2 && side2 < side3 {
+		return side1 + side2
+	}
+	if side1 < side3 && side3 < side2 {
+		return side1 + side3
+	}
+	return side2 + side3
+}
```
nobe4: make this a function of the structure `triangle`

11_structures/main.go
```diff
@@ -69,41 +73,327 @@ instances inside the function.
...
+
+func sumLengthSmallerLongestSide(side1, side2, side3 float64) bool {
+	sum := sumTwoShortestSides(side1, side2, side3)
+	biggest := longestSide(side1, side2, side3)
+	return sum <= biggest
+}
```
nobe4: make this a function of the structure `triangle`

11_structures/main.go
```diff
@@ -69,41 +73,327 @@ instances inside the function.
...
+// TODO: Create a function `compare` on the structure `triangle` that takes another
+// triangle as an argument and returns:
+// - -1 if the area of the first triangle is smaller than the area of the second
+// - 0 if the area of the first triangle is equal to the area of the second
+// - 1 if the area of the first triangle is greater than the area of the second
+// Hint: func (t1 triangle) compare(t2 triangle) int { ... }
+
+func semiParameter(side1, side2, side3 float64) float64 {
+	return (side1 + side2 + side3) / 0.5
+}
```
nobe4: make this a function of the structure `triangle`

11_structures/main.go
```diff
@@ -69,41 +73,327 @@ instances inside the function.
...
+
+// calculate square root
+// func sqrt(x float64) float64 {
+// }
+// squareRoot :=  math.Sqrt()
+
+func calcTriangleArea(semiParameter, side1, side2, side3 float64) float64 {
+	return math.Sqrt(semiParameter * (semiParameter * side1) * (semiParameter * side2) * (semiParameter * side3))
+}
```
nobe4: make this a function of the structure `triangle`

11_structures/main.go
riraum: Done

11_structures/main.go
riraum: Done

11_structures/main.go
riraum: Done

11_structures/main.go
riraum: Done

11_structures/main.go
riraum: Done

11_structures/main.go
```diff
@@ -157,25 +157,22 @@ func (t triangle) sumLengthSmallerLongestSide() bool {
...
-	semiParametert1 := semiParameter(t1.side1, t1.side2, t1.side3)
-	calcTriangleAreat1 := calcTriangleArea(semiParametert1, t1.side1, t1.side2, t1.side3)
-
-	semiParametert2 := semiParameter(t2.side1, t2.side2, t2.side3)
-	calcTriangleAreat2 := calcTriangleArea(semiParametert2, t2.side1, t2.side2, t2.side3)
+	calcTriangleAreat1 := t1.calcTriangleArea()
+	calcTriangleAreat2 := t2.calcTriangleArea()
 	if calcTriangleAreat1 < calcTriangleAreat2 {
 		return -1
 	}
```
riraum: @nobe4 What's the best practice, declare initial variables instead of repeatedly calling the functions?
I find both to be quite readable, depending on the variable/function naming.

11_structures/main.go
```diff
@@ -69,41 +73,323 @@ instances inside the function.
...
+}
+
+// TODO: Create a function `compare` on the structure `triangle` that takes another
+// triangle as an argument and returns:
+// - -1 if the area of the first triangle is smaller than the area of the second
+// - 0 if the area of the first triangle is equal to the area of the second
+// - 1 if the area of the first triangle is greater than the area of the second
+// Hint: func (t1 triangle) compare(t2 triangle) int { ... }
+
+func (t triangle) semiParameter() float64 {
```
nobe4: If I trust Wikipedia, it's actually called [Semiperimeter](https://en.wikipedia.org/wiki/Semiperimeter) :)

11_structures/main.go
nobe4: I'm not sure I understand, can you add examples of both ways in this thread?
FWIW, what is currently written in the file is fine :)

11_structures/main.go
riraum: I held off on renaming it semiPerimeterParameter, not sure if that would've been correct technically. 
Changed!

11_structures/main.go
riraum:
```
func (t triangle) calcTriangleArea() float64 {
	return math.Sqrt(t.semiPerimeter() * (t.semiPerimeter() * t.side1) * (t.semiPerimeter() * t.side2) * (t.semiPerimeter() * t.side3))
}

func (t1 triangle) compare(t2 triangle) int {
	calcTriangleAreat1 := t1.calcTriangleArea()
	calcTriangleAreat2 := t2.calcTriangleArea()
	if t1.calcTriangleArea() < t2.calcTriangleArea() {
		return -1
	}
	if t1.calcTriangleArea() == t2.calcTriangleArea() {
		return 0
	}
	if t1.calcTriangleArea() > t2.calcTriangleArea() {
		return 1
	}
	// debug, to remove
	return 6060
}
```
vs.

```
func (t triangle) calcTriangleArea() float64 {
semiPerimeter := t.semiPerimeter()
	return math.Sqrt(semiPerimeter * semiPerimeter * t.side1) * (semiPerimeter * t.side2) * (semiPerimeter * t.side3))
}

func (t1 triangle) compare(t2 triangle) int {
	calcTriangleAreat1 := t1.calcTriangleArea()
	calcTriangleAreat2 := t2.calcTriangleArea()
	if calcTriangleAreat1 < calcTriangleAreat2 {
		return -1
	}
	if calcTriangleAreat1 == calcTriangleAreat2 {
		return 0
	}
	if calcTriangleAreat1 > calcTriangleAreat2 {
		return 1
	}
	// debug, to remove
	return 6060
}
```
I did a mix of both.

nobe4: Ah I see! In most case, if you use the result of a function more than once, it's better to save it in a variable first. So your first example :)

11_structures/main.go
nobe4: I would also just call that function `semiPerimeter`. I don't see what `Parameter` would add here.

11_structures/main.go
riraum, line 1: @nobe4 Should I have the function accept an array/slice?
I had it working before when I had the loop in the main function.
But I guess from reading the instructions, it should be in the function `advIntro` instead.

Do I have to use to use the for/range syntax then to be able to loop? As currently I'm getting `invalid receiver type []person`

11_structures/main.go
```diff
@@ -69,44 +73,374 @@ instances inside the function.
...
+
+// TODO: Create a function that takes a list of person and prints "Hi, my
+// name is <name>, my BMI is <bmi> and I am <an adult|an infant>".
+func (p []person) advIntro() {
+	for i := 0; i < len(p); i++ {
+		name := p[i].name
+		bmi := p[i].bmi()
+		adultOrChild := p[i].adultOrChild()
+		fmt.Println("Hi, my name is", name, ", my BMI is", bmi, "and I am an", adultOrChild)
+	}
```
riraum: @nobe4 Should I have the function accept an array/slice?
I had it working before when I had the loop in the main function.
But I guess from reading the instructions, it should be in the function advIntro instead.

Do I have to use to use the for/range syntax then to be able to loop? As currently I'm getting `invalid receiver type []person`

riraum: Nvm. Had to change the function logic back to array/slice obviously. Was so fixated on the struct format.
Still have to figure out what the double points in `advIntro(randomArray[:])` do now.
Had the syntax changed to that by selecting the array.

riraum: It apparently converts it to a slice if it was an array and the uses all elements of it in the function call.
Removed that and the specific array size.

11_structures/main.go
```diff
@@ -69,58 +73,518 @@ instances inside the function.
...
+	fmt.Println(
+		"Print specific album year:",
+		zeppelin.year,
+	)
+	// fmt.Println(
+	// 	"Print specific artistArray album year:",
+	// 	artistArray[0].albums, map[int]album{2020: zeppelin.year},
+	// )
+	getFirstandLastAlbum(artistArray)
 }
```
riraum: @nobe4 I've been trying to find a way to target the year of map of albums of the artistArray. No luck so far.
The syntax for exactly that, is somehow confusing to me right now.
Where should I look?

nobe4: Have a look at https://go.dev/tour/moretypes/16

riraum: I read that, experimented a bit, but am not sure yet how to compare the album maps (or rather album years?) in the loop.
So far I wasn't able to define the correct syntax for my variable `firstAlbumYear`.
Do I have to use `reflect.DeepEqual` or something like that?

nobe4: Each artist has a map of `year => album`, if you loop with `range`, you will get both values for each entry in the map. You can then compare `year` with a `earliestYear` and `latestYear` and update variables accordingly. The comparison is between two `int`s

nobe4: This means that you have to do 2 loops: one that goes through each artist, and one that goes through each album for each artist.

riraum: I believe I have a working logic now.

11_structures/main.go
```diff
@@ -1,7 +1,11 @@
 package main
 
-/*
+import (
+	"fmt"
+	"math"
```
riraum, line 5: I did import `math` to be able to calculate the square root for the triangle area.
Had attempted to get it manually, but was a bit time consuming and I think it will also just be an approximate value.

nobe4: That's valid :) I forgot that you would need it, good job on your part finding how to do it.

11_structures/main.go
```diff
@@ -69,58 +73,522 @@ instances inside the function.
...
+
+// calculate square root
+// func sqrt(x float64) float64 {
+// }
+// squareRoot :=  math.Sqrt()
+
+func (t triangle) calcTriangleArea() float64 {
+	semiPerimeter := t.semiPerimeter()
+	return math.Sqrt(semiPerimeter * (semiPerimeter * t.side1) * (semiPerimeter * t.side2) * (semiPerimeter * t.side3))
+}
```
nobe4: couple of comments here:
- `triangle` in the name is redundant with the type
- `calc` is not needed, because there's nothing else you could do with the function in this context
```suggestion
func (t triangle) area() float64 {
	semiPerimeter := t.semiPerimeter()
	return math.Sqrt(semiPerimeter * (semiPerimeter * t.side1) * (semiPerimeter * t.side2) * (semiPerimeter * t.side3))
}
```

11_structures/main.go
```diff
@@ -69,58 +73,522 @@ instances inside the function.
...
+
+// TODO: Create a function on the structure `area` that returns a `float64`.
+type area struct {
+	side float64
+}
+
+func (a area) areaSize() float64 {
+	return a.side * a.side
+}
```
nobe4: I think this is not needed.
I meant: "in the structure `triangle` write a function `area` ....", which you've done later :)

11_structures/main.go
```diff
@@ -69,58 +73,522 @@ instances inside the function.
...
+
+func (t triangle) sumTwoShortestSides() float64 {
+	if t.side1 < t.side2 && t.side2 < t.side3 {
+		return t.side1 + t.side2
+	}
+	if t.side1 < t.side3 && t.side3 < t.side2 {
+		return t.side1 + t.side3
+	}
+	return t.side2 + t.side3
```
nobe4: Just for the fun of it, can you find a way to express this without doing any comparison and by using `t.longuestSide()` instead?

nobe4 (approved):
Approved with a few improvement comments, the code is otherwise good!

11_structures/main.go
riraum: Removed.

11_structures/main.go
riraum: Like this? Thought of doing a loop, but I think this is the simplest I can come up with.

11_structures/main.go
riraum: Makes sense!

11_structures/main.go
nobe4: yup, that's exactly what I had in mind :)

nobe4 (approved):

PR #17, opened 2024-11-05
13_packages

riraum:


riraum:
@nobe4 `Error: 13_packages/main.go:59:2: found packages geometry (geometry.go) and triangle (triangle.go) in /home/runner/work/forward/forward/internal/geometry
Error: Process completed with exit code 1.`
Do I have to specifically import both packages into `main.go`?
I tried to figure it out based on the instructions and lint message, but not sure what to do so far.

Apparently there shouldn't be multiple packages in the same directory...

Preliminarily fixed it, have to recheck the import of the respective packages inside each other later, will probably have to adjust paths.

riraum:
@nobe4 When I run `13_packages/main.go` I'm currently getting `# github.com/riraum/forward/internal/point
internal/point/point.go:6:2: "github.com/riraum/forward/internal/geometry" imported and not used
internal/point/point.go:14:9: undefined: float64Eq`
I do import `geometry` as I need `float64Eq` in `point`, but somehow it is imported but can't be used?

nobe4:
re-read the part on public/private functions in the main file. Moving things into different package is valid, but it means that private functions from other packages can't be accessed.

riraum:
@nobe4 I do understand now that packages can consist of multiple files in the respective directory, took me a while.

I would like to import the new packages in order to be able to debug and see if I did things right. Is that possible without adding it to the `master` branch? I tried to import it with the URL of this PR, but seems that I would need the path to `go.mod`.

nobe4:
@riraum yes, packages can be multiple files in a folder that have the same `package X` header in the files. Go will actually complain if you have more than one `package X` in a folder.

To import a new package, you can reference it with its full name `github.com/riraum/forward/...` and it will use the local files, no need to push to master first.

E.g. try adding `"github.com/riraum/forward/internal/person"` in the main file for this exercise.

internal/geometry/triangle.go
```diff
@@ -0,0 +1,86 @@
...
+	return (fullA + fullB + fullC) / 0.5
+}
+
+// TO CHECK
+func (t Triangle) Area() float64 {
+	fullA := t.addAXY()
+	fullB := t.addBXY()
+	fullC := t.addCXY()
+	semiPerimeter := t.semiPerimeter()
+	return math.Sqrt(semiPerimeter * (semiPerimeter * fullA) * (semiPerimeter * fullB) * (semiPerimeter * fullC))
```
riraum: @nobe4 I don't understand why I get the area of 0 for `demoTriangle` which has an A `Point` with a `X/Y` value of 0.
Seems to work for `demoTriangleY` which has an A`Point` where on coordinate is not 0.
Am I missing something here?

internal/geometry/triangle.go
```diff
@@ -0,0 +1,86 @@
...
+	// perimeter := t.Perimeter()
+	// Return data in req formatting
+	// return json.Marshal("Triangle{A: %s, B: %s, C: %s, Area: %s, Perimeter: %s,", t.A, t.B, t.C, t.Area(), t.Perimeter()})
+	// out, err := json.Marshal(t)
+	// if err != nil {
+	// 	panic(err)
+	// }
+	// fmt.Println(string(out))
+	return fmt.Sprintf("Triangle{\nA: %v, \nB: %v, \nC: %v, \nArea: %v, \nPerimeter: %v,", t.A, t.B, t.C, t.Area(), t.Perimeter())
+}
```
riraum: @nobe4 Am I on the right track here with the output formatting?
As it did say `Triangle` type I understood it to mean that I have to change the output format and can't create another struct (as I can't modify my existing definition).

I did read about `json.Marshal` and this simpler way I'm trying now.

internal/geometry/triangle.go
nobe4: Assuming you're using Heron's formula, your writing is incorrect: https://en.wikipedia.org/wiki/Heron%27s_formula

internal/geometry/triangle.go
nobe4: Yup that's what I had in mind :) No need to bother with `json` for now.

riraum:
@nobe4 I believe all is ready for review.
Not sure if you want me to clean up more debug statements, non used code and the like before or after your review.
I'm leaning towards after, as you might suggest changes.

13_packages/main.go
```diff
@@ -57,6 +57,10 @@ import (
 	"fmt"
 
 	"github.com/riraum/forward/internal/geometry"
+
+	"github.com/riraum/forward/internal/person"
+
+	"github.com/MakeNowJust/heredoc/v2"
```
nobe4, line 63: now that we manage imports as well, have a look at https://pkg.go.dev/golang.org/x/tools/cmd/goimports

Maybe can you change https://github.com/riraum/forward/blob/master/.github/workflows/ci.yaml to run `goimports` instead/alongside `go fmt` ?

nobe4: Otherwise I can do it, let me know :)

internal/person/person.go
```diff
@@ -0,0 +1,156 @@
...
+
+// TODO: Add a method to the 'Person' type called 'IsAdult' that returns a
+// boolean indicating if the person is an adult (i.e. 18 years or older).
+
+func (p Person) IsAdult() bool {
+	if p.Age >= 18 {
+		return true
+	} else {
+		return false
+	}
```
nobe4: How can you simplify this?

internal/person/person.go
```diff
@@ -0,0 +1,156 @@
...
+		// }
+	}
+
+	for _, value := range p.People {
+		sumHeight += int(value.Height)
+	}
+
+	for _, value := range p.People {
+		sumWeight += int(value.Weight)
+	}
```
nobe4: How can you simplify this?

internal/person/person.go
```diff
@@ -0,0 +1,156 @@
...
+		sumAge += value.Age
+		// debug
+		// fmt.Println("Print sumAge", sumAge)
+		// for _, sumActualAge := range sumAge.Age {
+		// 	sumAge += Person{}
+		// }
+	}
+
+	for _, value := range p.People {
+		sumHeight += int(value.Height)
```
nobe4: why do you convert to an int?

internal/person/person.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	for _, value := range p.People {
+		sumWeight += int(value.Weight)
```
nobe4: why do you convert to an int?

nobe4 (approved):
Couple of comments, and you can remove all the unneeded code/comments :)

internal/person/person.go
riraum: `amount` starts as`int`, as I have to divide the avg values with `amount` I do have to convert some values. Changed it so that I convert `amount` to `float64` and just convert it for `sumAge` once.
Your point was probably more that the `sum` should be a `float64` to begin with 😬

internal/person/person.go
riraum: See above.

internal/person/person.go
riraum: Yeah, now I feel stupid, haha. Done.

internal/person/person.go
riraum: Refactored!
I felt dirty writing it like this, but somehow the syntax didn't come to me then.

13_packages/main.go
riraum: Attempt done.

internal/geometry/triangle.go
```diff
@@ -0,0 +1,76 @@
...
+	return t.A.X + t.A.Y
+}
+
+func (t Triangle) addBXY() float64 {
+	return t.B.X + t.B.Y
+}
+
+func (t Triangle) addCXY() float64 {
+	return t.C.X + t.C.Y
+}
```
riraum: @nobe4 Okay like this, or should I try to simplify?
Feels a bit of a bit repetitive, but also can't currently come up with something more simple.

internal/person/person.go
nobe4: > Your point was probably more that the sum should be a float64 to begin with 😬

yup :)

internal/person/person.go
```diff
@@ -0,0 +1,119 @@
...
+func (p Person) IsAdult() bool {
+	return p.Age >= 18
+}
+
+// TODO: In the package 'person', create a new type called 'People' with
+// a field 'People' of type '[]Person'.
+
+type People struct {
+	People []Person
+}
```
nobe4: You an simplify this: 
```suggestion
type People []Person
```

You will need to change the code where you use `p People` to use `p` instead of `p.People`.

internal/person/person.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	for _, value := range namesSlice {
+		for _, secondValue := range namesSlice {
+			if value == secondValue {
+				continue
+			}
+			result = append(result, fmt.Sprintf("Hi %v from %v!\n", value, secondValue))
+		}
+	}
```
nobe4:
```suggestion
	for _, firstName := range namesSlice {
		for _, secondName := range namesSlice {
			if firstName == secondName {
				continue
			}
			result = append(result, fmt.Sprintf("Hi %v from %v!\n", firstName, secondName))
		}
	}
```

internal/geometry/triangle.go
nobe4: Agreed, especially since you always use the 3 values together.

How about a function that returns the 3 sides. Something like:

```go
func (t Triangle) addXY() (float64, float64, float64) {
	return t.A.X + t.A.Y, t.B.X + t.B.Y, t.C.X + t.C.Y
}
```

.github/workflows/ci.yaml
```diff
@@ -1,6 +1,8 @@
 # IGNORE ME
 name: CI
 
+import: "goimports"
+
```
nobe4:
```suggestion
```

nobe4: You need to setup a new step that install `goimport`

e.g.
```yaml
- name: Install goimports
  run: go install golang.org/x/tools/cmd/goimports@latest
```

.github/workflows/ci.yaml
```diff
@@ -35,6 +37,7 @@ jobs:
             }
 
             assert-nothing-changed go fmt ./...
+            assert-nothing-changed go goimports ./...
```
nobe4:
```suggestion
            assert-nothing-changed goimports ./...
```

13_packages/main.go
nobe4: Left some comments, let's continue the discussion there.

.github/workflows/ci.yaml
riraum: Done

.github/workflows/ci.yaml
riraum: Done

internal/person/person.go
riraum: Done!

internal/person/person.go
riraum: Done!

internal/geometry/triangle.go
riraum: Actually for `Area()` I need a specific side each. I was thinking of doing 
`func (t Triangle) addXY() float64{
	return t.A.X + t.A.Y + t.B.X + t.B.Y + t.C.X + t.C.Y
}` 
instead before, as else I would have to do a loop or something to get to one `float64` value that can divide. Then I realized that I need the specific sides. 
I could do an array and use the respective array index for `Area()`. Do you think worth the effort or leave it like it is after all?

.github/workflows/ci.yaml
riraum: Actually need more time to understand the syntax...

nobe4: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions

internal/geometry/triangle.go
nobe4: look closer at the function I wrote, it's returning each side independently

> I could do an array

would also work but offers little to what I suggested

.github/workflows/ci.yaml
riraum: I did add it and it seems correct I think. 
I went back and fort between my original changes, yours and now it seems to finally work.

internal/geometry/triangle.go
riraum: Added an improvement.
I do feel that declaring the sides as a variable once would be nicer, but maybe it's enough for now.

.github/workflows/ci.yaml
```diff
@@ -19,30 +19,8 @@ jobs:
...
-        run: gem install rubocop
-
-      # Javascript
-      - name: Set up Javascript
-        uses: actions/setup-node@v4
-        with:
-          node-version-file: ".node-version"
-
-      - name: Install Prettier
-        run: npm install prettier
```
nobe4: This removes what's on the main branch, please don't do that. You should keep both

riraum: My bad. Should be back. It is a bit challenging to navigate the changes on Git, add the right ones...

nobe4 (approved):

PR #19, opened 2024-11-09
14_tests

riraum:


riraum:
@nobe4 I'm finding it quite difficult, even after reading Go documentation, gobyexample on testing and the instructions and existing tests, to find the right syntax for the `Triangle` tests.
I think I have to do `want float64, float64, float64` as I will also get 3 return values no?

.github/workflows/ci.yaml
```diff
@@ -19,30 +19,8 @@ jobs:
...
-        run: gem install rubocop
-
-      # Javascript
-      - name: Set up Javascript
-        uses: actions/setup-node@v4
-        with:
-          node-version-file: ".node-version"
-
-      - name: Install Prettier
-        run: npm install prettier
```
nobe4: Don't remove those

internal/geometry/triangle_test.go
```diff
@@ -0,0 +1,22 @@
+package geometry
+
+import "testing"
+
+func TestAddXY(t *testing.T) {
+	tests := []struct {
+		ta   Triangle
+		want float64, float64, float64
```
nobe4: You can do it in a couple of ways:

```go
want1 float64
want2 float64
want3 float64
```

```go
want []float64
```

.github/workflows/ci.yaml
riraum: I was waiting to merge from master again, once PR #17 was complete. Will do of course.

internal/person/person_test.go
riraum, line 1: @nobe4 I did find the syntax for `person` quite a bit easier.
Apart from `TestPeopleGreet` where I still have to figure out how to deal with a slice, they seem to run at least.
I guess I could convert the slice output into one string one.
Are the test usable like this?

internal/geometry/triangle_test.go
riraum, line 1: @nobe4 I read and experimented a bit more, but still seem to not really get the syntax for the `Triangle` somehow. 
Am I on the right track at least?

nobe4: You are, couple of comments:

> missing type in composite literal

means you need to add the type of the structure before the `{`.
E.g. `{0, 0},` => `Point{0, 0},`

Also, because you're using an array of structure, you need to wrap things a bit differently:

```go
tests := []struct{
  ta Triangle
  // ...
}{
  { // first item of the list
     ta: Triangle{ /* ... */ },
     // ...
  },
  { // second item of the list
     ta: Triangle{ /* ... */ },
     // ...
  },
}

```

nobe4: Read carefuly https://github.com/riraum/forward/blob/master/internal/geometry/point_test.go#L8-L24

internal/person/person_test.go
nobe4: Yes, `greet` is quite simple, so a single person might be enough.

For other methods, make sure to test with multiple persons.

internal/geometry/triangle_test.go
```diff
@@ -27,12 +27,32 @@ func TestAddXY(t *testing.T) {
...
+		// },
+		{ta: Triangle{
+			A: Point{X: 0, Y: 0},
+			B: Point{X: 0, Y: 4},
+			C: Point{X: 3, Y: 0},
+		},
+			want1: 0,
+			want2: 4,
+			want3: 3,
+		},
```
nobe4: indent suggestion:
```suggestion
		{
		  ta: Triangle{
			  A: Point{X: 0, Y: 0},
			  B: Point{X: 0, Y: 4},
			  C: Point{X: 3, Y: 0},
		  },
			want1: 0,
			want2: 4,
			want3: 3,
		},
```

internal/person/person_test.go
```diff
@@ -0,0 +1,171 @@
...
+	test := anna.BMI()
+	wantedResult := 19.140625
+	if test != wantedResult {
+
+		t.Errorf("anna.BMI() = %v; want %v", test, wantedResult)
+	}
+}
+
+func TestIsAdult(t *testing.T) {
+	anna := Person{
```
nobe4: Make more than 1 test: test on < 18, = 18, > 18

internal/person/person_test.go
```diff
@@ -0,0 +1,171 @@
...
+		Age:    17,
+		Height: 160,
+		Weight: 49,
+	}
+	test := anna.BMI()
+	wantedResult := 19.140625
+	if test != wantedResult {
+
+		t.Errorf("anna.BMI() = %v; want %v", test, wantedResult)
+	}
```
nobe4: Also do a couple of different BMI here

internal/geometry/triangle_test.go
riraum: Why?
It's unclear to me why the triangle values/keys should have more spacing than just a tab. Also why the `want` values/keys should have a lot of spacing, but not sit at on the same position as the values/keys above?
Should I change my spacing/indentation settings?

internal/person/person_test.go
riraum: Added

internal/person/person_test.go
riraum: Added.

internal/person/person_test.go
```diff
@@ -0,0 +1,245 @@
...
+	// 	"Hi Jan from Luna!",
+	// 	"Hi Jan from Tay!",
+	// 	"Hi Tay from Anna!",
+	// 	"Hi Tay from Luna!",
+	// 	"Hi Tay from Jan!"}
+	// if test != wantedResult {
+
+		t.Errorf("team.Greet() = %v; want %v", test, wantedResult)
+	}
+}
```
riraum: @nobe4 Unsure what to do for this one. I seem to not be able to just fill the slice manually, as it apparently can only be compared to nil.
Do I loop then?
Just format the the results as a string in array syntax?
I could run `gotests` again and see what it would do, but feels a bit like cheating.

nobe4: That's true you cannot compare slices directly, but you can try using https://pkg.go.dev/slices#Equal

e.g. https://go.dev/play/p/Jde00Lv27wv

internal/geometry/triangle_test.go
nobe4: The way I see it is:

```go
var x = {
  a: 1,
  b: 2,
}

// or

var y = { a: 1, b: 2 } 
```

But mixing the two is weird.

e.g.

```go
var z = { a: 1, b: 2,
}
```

internal/person/person_test.go
nobe4: You should switch the test to use an array of struct:

```go
tests := []struct {
  p Person
  want bool
} {
  // ...
}
```

internal/person/person_test.go
```diff
@@ -0,0 +1,245 @@
...
+	}
+	test := team.Greet()
+	namesSlice := []string{}
+	wantedResult := []string{}
+	for _, firstName := range team {
+		namesSlice = append(namesSlice, firstName.Name)
+	}
+
+	for _, value := range namesSlice {
+		for_, familyName := range namesSlice {
```
nobe4:
```suggestion
		for _, familyName := range namesSlice {
```

internal/person/person_test.go
```diff
@@ -0,0 +1,245 @@
...
+	test := anna.Greet()
+	wantedResult := "Hello Anna"
+	if test != wantedResult {
+
+		t.Errorf("anna.Greet() = %v; want %v", test, wantedResult)
+	}
+}
+
+func TestBMI(t *testing.T) {
+	anna := Person{
```
nobe4: You should switch the test to use an array of struct:

```go
tests := []struct {
  p Person
  want float64
} {
  // ...
}
```

internal/geometry/triangle_test.go
riraum: Ah, to move to a new line in L34 after opening `{`, that makes sense!

internal/person/person_test.go
riraum: Ouch, didn't spot that, thanks!

internal/person/person_test.go
riraum: Attempt done.

internal/person/person_test.go
riraum: Done.

internal/person/person_test.go
riraum: Done.

riraum:
@nobe4 Seems I have achieved 100% test coverage somehow.

internal/person/person_test.go
```diff
@@ -0,0 +1,245 @@
...
+			want: 19.723865877712033,
+		},
+		{
+			p: Person{
+				Name:   "Jan",
+				Age:    29,
+				Height: 188,
+				Weight: 90,
+			},
+			want: 25.464010864644635},
```
nobe4:
```suggestion
			want: 25.464010864644635,
		},
```

internal/person/person_test.go
```diff
@@ -0,0 +1,245 @@
...
+			want: true,
+		},
+		{
+			p: Person{
+				Name:   "Tay",
+				Age:    18,
+				Height: 190,
+				Weight: 87,
+			},
+			want: true},
```
nobe4:
```suggestion
			want: true,
		},
```

internal/person/person_test.go
```diff
@@ -0,0 +1,245 @@
...
+	wantedResult := team[3]
+	if test != wantedResult {
+
+		t.Errorf("team.Oldest() = %v; want %v", test, wantedResult)
+	}
+}
+
+func TestPeopleGreet(t *testing.T) {
+	team := People{
+		{Name: "Anna",
```
nobe4:
```suggestion
		{
			Name: "Anna",
```

internal/person/person_test.go
```diff
@@ -0,0 +1,245 @@
...
+		Weight: 68.5,
+	}
+	if test != wantedResult {
+		t.Errorf("team.Average() = %v; want %v", test, wantedResult)
+	}
+}
+
+func TestOldest(t *testing.T) {
+	team := People{
+		{Name: "Anna",
```
nobe4:
```suggestion
		{
			Name: "Anna",
```

internal/person/person_test.go
```diff
@@ -0,0 +1,245 @@
...
+		got := test.p.IsAdult()
+		if got != test.want {
+			t.Errorf("IsAdult(%v) = %v, want %v", test.p, got, test.want)
+		}
+	}
+}
+
+func TestAverage(t *testing.T) {
+	team := People{
+		{Name: "Anna",
```
nobe4:
```suggestion
		{
			Name: "Anna",
```

nobe4 (approved):
Small formatting improvements, otherwise LGTM

internal/person/person_test.go
riraum: Done

internal/person/person_test.go
riraum: Done

internal/person/person_test.go
riraum: Done

internal/person/person_test.go
riraum: Done

internal/person/person_test.go
riraum: Done

PR #23, opened 2024-11-13
14_tests: Add slight testing improvement for TestAddXY

riraum:
Saw some quick way to make the test a bit better.


nobe4 (commented):
You should be able to use an || between each condition instead, that would make for a single if

riraum:
Like this?
Wow, less lines :)

nobe4 (approved):

PR #24, opened 2024-11-14
15_tictactoe

riraum:


15_tictactoe/go/grid/grid.go
```diff
@@ -1,10 +1,16 @@
...
 }
 
 func (g Grid) String() string {
 	// TODO
+	// |||____
+	// |__|__|__|
+	// |__|__|__|
+	// |__|__|__|
 	return ""
 }
```
riraum: @nobe4 Finding it a bit hard to start. Is this very basic idea going in the right direction?

nobe4: Try using a single list of ints for the cell:

```go
type Grid struct {
  Cells []int
}
```

riraum: @nobe4 Could I get more hints for after that?

15_tictactoe/go/grid/grid.go
```diff
@@ -1,10 +1,26 @@
...
+
 func (g Grid) String() string {
 	// TODO
+	// |||____
+	// |__|__|__|
+	// |__|__|__|
+	// |__|__|__|
+	// row := g.Cells * "__"
+	// column := g.Cells
+	Cells := []string{}
```
nobe4, line 20: you can do this in a single operation, using https://pkg.go.dev/fmt#Sprintf

The function should look like

```go
func (g Grid) String() string {
  return fmt.Sprintf(
    "<TODO>",
    args...
  )
}
```

15_tictactoe/go/grid/grid.go
nobe4: you can do this in a single operation, using https://pkg.go.dev/fmt#Sprintf

The function should look like

```go
func (g Grid) String() string {
  return fmt.Sprintf(
    "<TODO>",
    // g.Cells...
  )
}
```

riraum: @nobe4 Trying to wrap my head around the logic again. Do I have it now with the last commit?

nobe4: 8b5035b doesn't print the grid as a grid.
Have a read at https://pkg.go.dev/fmt, especially the https://pkg.go.dev/fmt#example-package-Formats to see how you can format the output into a table.

E.g. output

```
+---+---+---+
| X |   |   |
+---+---+---+
|   | O | O |
+---+---+---+
| X |   |   |
+---+---+---+
```

riraum: @nobe4 Another attempt. If that's not it, I think I have to try to figure out how to break it down further and manage to run and test the code first.

riraum: @nobe4 Added string grid logic for output first with int values then with string output. I guess string output would be better?

riraum: @nobe4 have
- visual string output
- win output

Trying to get the indexes of the free cell slice returned now.
Suppose I have to loop with range for that?
Did it before, but then I ignored the indexes and just returned the values. Seems a bit more tricky to just return the indexes...

15_tictactoe/go/grid/grid.go
```diff
@@ -1,10 +1,140 @@
...
+	// cell := "|__|"
+	// count := g.Cells
+	return fmt.Sprintf(`
++---+---+---+
+| %+v | %+v | %+v |
++---+---+---+
+| %+v | %+v | %+v |
++---+---+---+
+| %+v | %+v | %+v |
++---+---+---+`, g.Cells[2], g.Cells[1], g.Cells[2], g.Cells[3], g.Cells[4], g.Cells[5], g.Cells[6], g.Cells[7], g.Cells[8],
```
nobe4:
```suggestion
+---+---+---+`, g.Cells[0], g.Cells[1], g.Cells[2], g.Cells[3], g.Cells[4], g.Cells[5], g.Cells[6], g.Cells[7], g.Cells[8],
```

nobe4 (approved):

15_tictactoe/go/grid/grid.go
riraum: Done

riraum:
@nobe4 On the last TestFreeCells one.
Am I assuming correctly that I should use:
- Grid slices with cell values
- slices.Equal to compare the grid slices values with the wantedResult slices?

Having slice structs with wanted values doesn't seem to work as I can't compare slices directly.
Hence the need to use slices.Equal?

nobe4:
Using slices.equal and store the expected slice in the test struct sounds like the best approach :)

15_tictactoe/go/grid/grid_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+		t.Errorf("testGrid.String() = %v; want %v", test, wantedResult)
+	}
+}
+
+func TestIsWin(t *testing.T) {
+	tests := []struct {
+		g     Grid
+		want1 bool
+		want2 string
```
nobe4: Can you use better names here? Like `wantIsWin` and `wantWinner` for example?

15_tictactoe/go/grid/grid.go
```diff
@@ -1,10 +1,140 @@
 package grid
 
+import (
+	"fmt"
+)
+
+// Define column and row
 type Grid struct {
-	// TODO
+	Cells [9]int
```
nobe4: Have a look at all your functions and think about how you want to represent each player. You currently have numbers and strings. I would choose a single type to represent players and simplify your code thusly.

15_tictactoe/go/grid/grid_test.go
riraum: Done

15_tictactoe/go/grid/grid.go
riraum: I think it should be strings, according to the instructions. I did run into issues when trying to use strings and therefore just switched to int for the meantime.
Will revisit after the the last test works.

riraum:
@nobe4 For the last `TestFreeCells` test.
I tried to create a `testGrid` struct in addition to my `wantedResult` struct.
I would then have to loop with range through `testGrid`, apply the function I want to test `FreeCells` and store the resulting indexes in a new struct.
Only then I can compare the resulting struct with my `wantedResult` struct through `slices.Equal` and not get a miss match due to different types etc.

Does that make sense?

15_tictactoe/go/grid/grid_test.go
```diff
@@ -0,0 +1,160 @@
...
+	// 		g: Grid{
+	// 			Cells: [9]int{0, 0, 0, 0, 0, 0, 0, 0, 0},
+	// 		},
+	// 		want: []int{0, 1, 2, 3, 4, 5, 6, 7, 8},
+	// 	},
+	// }
+
+	// testGrid struct to test multiple cases
+	testGrid := []struct {
+		g Grid
```
nobe4: Try to do this instead:
```suggestion
		g Grid
		wanted []int
```

riraum: I think I have done it!?

15_tictactoe/go/grid/grid.go
riraum: Changed to strings!

nobe4 (approved):

PR #25, opened 2025-01-06
15 tictactoe random

riraum:
Ready for review.

### DIFF
--- a/15_tictactoe/go/random/random.go
+++ b/15_tictactoe/go/random/random.go
@@ -1,6 +1,10 @@
 package random
 
+import (
+	"math/rand"
+)
+
 func Choose(values []int) int {
-	// FIXME
-	return 0
+	// Select a random value from slice of values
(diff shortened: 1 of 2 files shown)


15_tictactoe/go/random/random.go
```diff
@@ -1,6 +1,10 @@
...
+import (
+	"math/rand"
+)
+
 func Choose(values []int) int {
-	// FIXME
-	return 0
+	// Select a random value from slice of values
+	return values[rand.Intn(len(values))]
 }
```
riraum, line 10: @nobe4 If I understand correctly, I'm returning an `index` of the `values` `slice`.
The function `Intn` of package `rand` that was imported returns a a random `integer` based on the length of the `values` `slice`.
Correct?

nobe4: Yeah, that looks good.

nobe4 (approved):

PR #26, opened 2025-01-06
15_tictactoe - Add io logic

riraum:
Ready for review

### DIFF
--- a/15_tictactoe/go/io/io.go
+++ b/15_tictactoe/go/io/io.go
@@ -3,8 +3,8 @@ package io
 import "fmt"
 
 func Read(prompt string) string {
+	var input string
+	fmt.Scan(&input)
 	fmt.Printf("%s > ", prompt)
-
-	// FIXME
-	return ""
+	return input
 }
--- a/15_tictactoe/go/io/io.go
+++ b/15_tictactoe/go/io/io.go
@@ -4,7 +4,7 @@ import "fmt"
 
 func Read(prompt string) string {
 	var input string
-	fmt.Scan(&input)
 	fmt.Printf("%s > ", prompt)
+	fmt.Scan(&input)
 	return input
 }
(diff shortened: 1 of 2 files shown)


15_tictactoe/go/io/io.go
```diff
@@ -3,8 +3,8 @@ package io
 import "fmt"
 
 func Read(prompt string) string {
+	var input string
+	fmt.Scan(&input)
 	fmt.Printf("%s > ", prompt)
-
-	// FIXME
-	return ""
+	return input
```
riraum: @nobe4 The go documentation for fmt.Scan was not clear enough for me, found a Stackoverflow post.
I seem to need the `&` operator, to get the chance to actually get the prompt to input something.
Could not seem to find a good explanation for the function of it.

nobe4: It's a pointer, meaning the memory address of the variable.
It's used to pass variables around so that they can be written from different places.

Have a look at 
- https://go.dev/tour/moretypes/1
- https://en.wikipedia.org/wiki/Pointer_(computer_programming)#Go

riraum: Thanks, had a read. Also read this one:
https://stackoverflow.com/questions/20895552/how-can-i-read-from-standard-input-in-the-console
I still don't get why I need a pointer to be able to input something, but maybe it's not so important to understand that specific logic for now :)

nobe4: basically: the function will store the value inputed into the variable you pass, but in order to modify the variable inside the function, you need to use its address.

nobe4 (changes requested):
The display is inverted.
You should display the prompt before asking for the input.

riraum:
@nobe4 Should be fixed.

nobe4 (approved):

PR #27, opened 2025-01-06
15_tictactoe - Combine packages, make game run

riraum:
Put all the packages together.

### DIFF
--- a/15_tictactoe/go/game/game.go
+++ b/15_tictactoe/go/game/game.go
@@ -33,10 +33,10 @@ func Play() {
 	}
 
 	for i := 0; i <= len(g.Cells); i++ {
-		if i == len(g.Cells) {
-			fmt.Println("Grid full, no win!")
-			break
-		}
+		// if i == len(g.Cells) {
+		// 	fmt.Println("Grid full, no win!")
+		// 	break
+		// }
 		// Output empty grid to help understand the concept?
 		fmt.Println(g.String())
 		// Ask player to select a grid index
@@ -48,9 +48,14 @@ func Play() {
 		fmt.Println(g.String())
 		// Check if there is a winner
 		fmt.Println(g.IsWin())
-		if g.IsWin() == true, "X won" || g.IsWin() == true, "O won" {
+		// WIP break condition
+		win, _ := g.IsWin()
+		if win {
 			break
 		}
+		// if g.IsWin() == true, "X won" || g.IsWin() == true, "O won" {
+		// 	break
+		// }
 		// Get freecells
 		freeCellSlice := []int{}
 		freeCellSlice = append(freeCellSlice, g.FreeCells()...)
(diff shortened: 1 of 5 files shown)


15_tictactoe/go/game/game.go
```diff
@@ -8,7 +8,64 @@ import (
...
+		fmt.Println("You chose", PlayerMove)
+		// Add player marker to correct index of grid
+		g.Cells[PlayerMove] = "X"
+		// Output grid to visualize choice
+		fmt.Println(g.String())
+		// Check if there is a winner
+		fmt.Println(g.IsWin())
+		if g.IsWin() == true, "X won" || g.IsWin() == true, "O won" {
+			break
+		}
```
riraum: @nobe4 I think the break logic is my biggest issue right now.
The loop should break if `IsWin` is `true`.
I'm struggling to find the correct syntax for the `bool` and `string` response.

15_tictactoe/go/game/game.go
```diff
@@ -8,7 +8,64 @@ import (
...
+	// Initialize empty grid
+	g := grid.Grid{
+		Cells: [9]string{" ", " ", " ", " ", " ", " ", " ", " ", " "},
+	}
+
+	for i := 0; i <= len(g.Cells); i++ {
+		if i == len(g.Cells) {
+			fmt.Println("Grid full, no win!")
+			break
+		}
```
nobe4: This is redundant with the condition on line 35.

15_tictactoe/go/game/game.go
nobe4: `IsWin` returns 2 values, you need to store them before you can use them.

15_tictactoe/go/game/game.go
```diff
@@ -8,7 +8,64 @@ import (
...
+		// Output grid to visualize choice
+		fmt.Println(g.String())
+		// Check if there is a winner
+		fmt.Println(g.IsWin())
+		if g.IsWin() == true, "X won" || g.IsWin() == true, "O won" {
+			break
+		}
+		// Get freecells
+		freeCellSlice := []int{}
+		freeCellSlice = append(freeCellSlice, g.FreeCells()...)
```
nobe4: This can be done in a single line

15_tictactoe/go/game/game.go
riraum: Removed.

15_tictactoe/go/game/game.go
riraum: Could I get a hint on where to find the right syntax?

15_tictactoe/go/game/game.go
```diff
@@ -8,44 +8,72 @@ import (
...
+		// freeCellSlice = append(freeCellSlice, g.FreeCells()...)
+		freeCellSlice := g.FreeCells()
+		fmt.Println(freeCellSlice)
+		// Break condition in case of no free cell
+		freeCells := g.FreeCells()
+		if len(freeCells) == 0 {
+			break
+		}
+		// Make random computer choice, based on freecells
+		computerMove := random.Choose(freeCellSlice)
```
nobe4: You call `Freecells` twice, store it in two different variables. It should be a single one.

riraum: Removed!

15_tictactoe/go/game/game.go
riraum: Should be done.

nobe4 (approved):
Approving for now, have a look at https://github.com/riraum/forward/issues/28 for further improvements.

Issue #28, opened 2025-01-08
TicTacToe improvements


nobe4:
Various improvements to make the code even better

- [x] in `io.Read`, validate that the input is a valid integer
- [x] in `io.Read`, validate that the input is a valid choice (i.e. not an already chosen one)
- [x] in `grid`, create a function that initialize an empty grid
- [x] in `grid.String`, show `X` / `O` for chosen cells and their index otherwise, to help choosing
- [x] in `grid.IsWin`, you can refactor the function to simplify it
- [x] in `game.Play`, refactor the game loop to simplify it
  Hint: the loop should look like:
  ```go
  for i := 0; i <= 9; i++ {
    // print the grid

    if i % 2 == 0 {
      // player make a move
    } else {
      // computer make a move
    }

    // update the grid
    // check for victory
  } 
  ```
- [x] remove all unnecessary comments

PR #29, opened 2025-01-08
15_tictactoe - io input validation

riraum:
#28 
>  in io.Read, validate that the input is a valid integer
> in io.Read, validate that the input is a valid choice (i.e. not an already chosen one)

### DIFF
--- a/15_tictactoe/go/io/io.go
+++ b/15_tictactoe/go/io/io.go
@@ -9,40 +9,22 @@ Break loop if input is an integer, else keep looping
 Break loop if input is on a free cell, else keep looping
 */
 
-/*
-initialize function that can be used outside of this package
-set the input value the variable name `prompt` of the type string
-set the return value to type `int`
-*/
 func Read(prompt string) int {
-	// initialize variable `input` of type `int`
 	var input int
-	// start `for` loop, counting from 0, until <=3
 	for i := 0; i <= 3; i++ {
-		// print the variable value of `prompt`
+
 		fmt.Printf("%s > ", prompt)
-		// set variable, discarding the value, only focusing on a potential error of the input given by the user, stored in the variable input
 		_, err := fmt.Scan(&input)
-		// if the result of the function call to `valid` returns false
-		if !valid(input) {
-			// print variable `err`, the error message of the that resulted from the user input
-			fmt.Println(err)
-			// break the loop
-			break
-			// if that is not the case, the function call to `valid` returns true, continue the loop
-		} else {
+		if valid(input) {
 			continue
+		} else {
+			fmt.Println(err)
+			break
 		}
 	}
-	// in case the loop was not broken before, return the input
 	return input
 }
 
-/*
-Initialize the function `valid`, only usable in this package
-Name the value that is passed the variable `input` of type `int`
-Set the return value to type `bool`
-*/
 func valid(input int) bool {
 	// var result bool
 	// if err != nil {
@@ -51,11 +33,9 @@ func valid(input int) bool {
 	// if value == 00 {
 	// 	return true
 	// }
-	// if the `input` is >= 0 AND <=8 return false
 	if input >= 0 && input <= 8 {
 		return false
 	}
-	// if that's not the case, also return false
 	return false
 }
 
--- a/15_tictactoe/go/io/io.go
+++ b/15_tictactoe/go/io/io.go
@@ -14,8 +14,8 @@ func Read(prompt string) int {
 	for i := 0; i <= 3; i++ {
 
 		fmt.Printf("%s > ", prompt)
-		value, err := fmt.Scan(&input)
-		if valid(value, err) {
+		_, err := fmt.Scan(&input)
+		if valid(input) {
 			continue
 		} else {
 			fmt.Println(err)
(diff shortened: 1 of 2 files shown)


riraum:
@nobe4 Tried to import `game` into `io`.
Tried to run `PlayerMove := io.Read("What's your choice? (Enter index 0-8 if not taken before)")` with a second error variable.

15_tictactoe/go/io/io.go
```diff
@@ -2,9 +2,25 @@ package io
...
 	var input int
-	fmt.Printf("%s > ", prompt)
-	fmt.Scan(&input)
+	for i := 0; i <= 3; i++ {
+
+		fmt.Printf("%s > ", prompt)
+		value, err := fmt.Scan(&input)
+		if err != nil {
+			continue
+		}
```
nobe4: You should print what the error is, to help debug.

15_tictactoe/go/io/io.go
```diff
@@ -2,9 +2,25 @@ package io
...
+		if value == 0 || value == 1 || value == 2 || value == 3 || value == 4 || value == 5 || value == 6 || value == 7 || value == 8 {
```
nobe4:
```suggestion
		if value >= 0 && value <= 8 {
```

15_tictactoe/go/io/io.go
```diff
@@ -2,9 +2,25 @@ package io
...
+			break
+		} else {
+			break
+		}
```
nobe4: wait, so the if is 
```go
if condition {
  break
} else {
  break
}
```

Isn't that a bit useless?

15_tictactoe/go/io/io.go
nobe4: Alternatively, you can read a string and convert the string to an int with https://pkg.go.dev/strconv

15_tictactoe/go/io/io.go
riraum: I might have written that without too much thinking 😬
Removed the else break statement.

15_tictactoe/go/io/io.go
riraum: That looks way more decent. Typed it like this :)

15_tictactoe/go/io/io.go
riraum: Might have it working now.

15_tictactoe/go/io/io.go
nobe4: I think you ought to separate the function a bit:

- have the part that does the input and loop until it's valid
- extract the validation to another function, so you can write tests for it

riraum: Alright. Seem to also have to work on getting an error when something other than a double digit integer is entered.
That and figuring out how to check for free cells in this package.

15_tictactoe/go/io/io.go
```diff
@@ -2,9 +2,55 @@ package io
...
+
 func Read(prompt string) int {
 	var input int
-	fmt.Printf("%s > ", prompt)
-	fmt.Scan(&input)
+	for i := 0; i <= 3; i++ {
+
+		fmt.Printf("%s > ", prompt)
+		value, err := fmt.Scan(&input)
+		if valid(value, err) {
```
nobe4: That's not a correct pattern in go. You ought to check the error where it's defined.

The `value` is _not_ what you typed, but the number of byte stored. You probably can discard it.

If `err` here isn't nil, then something went wrong with `fmt.Scan`, not necessarily that the value is wrong.

riraum: I'm finding this quite hard, to separate the logic into 2 functions. I had to do it at least once before, was similarly challenging.

Did some adjustments and at least it seems to fail, with error message `out of range` when I enter 1 or several `9`s.

15_tictactoe/go/io/io.go
```diff
@@ -33,9 +51,11 @@ func valid(input int) bool {
 	// if value == 00 {
 	// 	return true
 	// }
+	// if the `input` is >= 0 AND <=8 return false
```
nobe4: e.g. from our conversation, this comment adds nothing to the code below.
Try to refrain from adding noise with such comments when the code is "simple".

riraum: Removed.

riraum:
@nobe4 Please review!

15_tictactoe/go/io/io_test.go
```diff
@@ -0,0 +1,31 @@
...
+// 	}
+// }
+
+func TestValid(t *testing.T) {
+	tests := []struct {
+		input string
+		want  bool
+	}{
+		{"3", true},
+		{"99", false},
```
nobe4, line 13: add a test for an empty string, for a non-number string as well

15_tictactoe/go/io/io_test.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+// func testRead(t *testing.T) {
+// 	tests := []struct {
+// 		input int
+// 		want int
+// 	}{
+// 		{}
+// 	}
+// }
```
nobe4: can this be removed?

nobe4 (approved):
Added an improvement suggestion, otherwise good to go :shipit:

15_tictactoe/go/io/io_test.go
riraum: Added!

15_tictactoe/go/io/io_test.go
riraum: Removed!

PR #30, opened 2025-01-20
15_tictactoe - io input validation duplicate cell check

riraum:
#28 
> io.Read, validate that the input is a valid choice (i.e. not an already chosen one)

### DIFF
--- a/15_tictactoe/go/io/io.go
+++ b/15_tictactoe/go/io/io.go
@@ -27,10 +27,8 @@ func valid(input string, freeCellSlice []int) bool {
 		return false
 	}
 
-	for _, value := range freeCellSlice {
-		if slices.Contains(freeCellSlice, value) {
-			return true
-		}
+	if slices.Contains(freeCellSlice, intInput) {
+		return true
 	}
 
 	fmt.Printf("'%v' is already taken! Try another choice\n", intInput)
--- a/15_tictactoe/go/io/io.go
+++ b/15_tictactoe/go/io/io.go
@@ -2,6 +2,7 @@ package io
 
 import (
 	"fmt"
+	"slices"
 	"strconv"
 )
 
@@ -27,7 +28,7 @@ func valid(input string, freeCellSlice []int) bool {
 	}
 
 	for _, value := range freeCellSlice {
-		if intInput == value {
+		if slices.Contains(freeCellSlice, value) {
 			return true
 		}
 	}
--- a/15_tictactoe/go/io/io.go
+++ b/15_tictactoe/go/io/io.go
@@ -5,16 +5,6 @@ import (
 	"strconv"
 )
 
-/*
-Input validation:
-Create loop
-Break loop if input is an integer, else keep looping
-Break loop if input is on a free cell, else keep looping
-*/
-
-// Logic to check if duplicate user input only
-// var takenCells []int
-
 func Read(prompt string, freeCellSlice []int) int {
 	var input string
 	for {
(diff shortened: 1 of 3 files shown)


riraum:
@nobe4 Could I get a preliminary review please?
Have to fix the test, remove debug code and comments, but the `io.Read` should check for duplicate user/computer choice now.

nobe4:
Looks promising, I would have done something similar :)

15_tictactoe/go/io/io.go
```diff
@@ -5,37 +5,33 @@ import (
...
 		return false
 	}
-	if intInput >= 0 && intInput <= 8 {
-		return true
+
+	for _, value := range freeCellSlice {
+		if intInput == value {
+			return true
+		}
 	}
```
nobe4: https://pkg.go.dev/slices#Contains

riraum: Implemented!

15_tictactoe/go/io/io.go
```diff
@@ -2,40 +2,35 @@ package io
...
 		return false
 	}
-	if intInput >= 0 && intInput <= 8 {
+
+	if slices.Contains(freeCellSlice, intInput) {
 		return true
 	}
 
-	fmt.Println("Print invalid input", intInput)
+	fmt.Printf("'%v' is already taken! Try another choice\n", intInput)
```
nobe4:
```suggestion
	fmt.Printf("'%d' is already taken! Try another choice\n", intInput)
```

`%v` works in most cases, but `%d` is better suited for integer/numbers: https://en.wikipedia.org/wiki/Printf#Type_field

15_tictactoe/go/io/io_test.go
```diff
@@ -6,17 +6,18 @@ import (
...
-		want  bool
+		input     string
+		testSlice []int
+		want      bool
 	}{
-		{"3", true},
-		{"99", false},
-		{"", false},
-		{"abc", false},
+		{"3", []int{3}, true},
```
nobe4, line 13:
```suggestion
		{"3", []int{3}, true},
		{"3", []int{4}, false},
```

15_tictactoe/go/io/io_test.go
```diff
@@ -6,17 +6,18 @@ import (
...
+		{"99", []int{}, false},
+		{"", []int{}, false},
+		{"abc", []int{}, false},
 	}
 
 	for _, test := range tests {
-		got := valid(test.input)
+		got := valid(test.input, []int{3})
```
nobe4:
```suggestion
		got := valid(test.input, test.testSlice)
```

15_tictactoe/go/io/io_test.go
```diff
@@ -6,17 +6,18 @@ import (
 
 func TestValid(t *testing.T) {
 	tests := []struct {
-		input string
-		want  bool
+		input     string
+		testSlice []int
```
nobe4: I would change the name to this, for something more descriptive.
```suggestion
		validChoices []int
```

15_tictactoe/go/io/io.go
riraum: Done. I guess go as precise as possible then.

15_tictactoe/go/io/io_test.go
riraum: Added.

15_tictactoe/go/io/io_test.go
riraum: Good call, done.

15_tictactoe/go/io/io_test.go
riraum: Makes sense. Added.

nobe4 (approved):

PR #33, opened 2025-01-21
15_tictactoe - Refactor isWin func to simplify it

riraum:
#28 > in grid.IsWin, you can refactor the function to simplify it

### DIFF
--- a/15_tictactoe/go/grid/grid.go
+++ b/15_tictactoe/go/grid/grid.go
@@ -29,6 +29,21 @@ func (g Grid) String() string {
 	)
 }
 
+func (g Grid) checkLine(x, y, z int) (bool, string) {
+	computerWin := "Computer wins!"
+	humanWin := "Human wins!"
+
+	if g.Cells[x] != g.Cells[y] || g.Cells[x] != g.Cells[z] {
+		return false, ""
+	}
+
+	if g.Cells[x] == "X" {
+		return true, humanWin
+	}
+
+	return true, computerWin
+}
+
 // Create variable for empty cell, filled with X and with O
 // Return row of cells based on slice length, target index of slice and assign one of the 3 cell types
 // Return column of cells based and do the same as with the row
--- a/15_tictactoe/go/grid/grid.go
+++ b/15_tictactoe/go/grid/grid.go
@@ -36,7 +36,7 @@ func (g Grid) String() string {
 | %+v | %+v | %+v |
 +---+---+---+
 | %+v | %+v | %+v |
-+---+---+---+`, g.Cells[0], g.Cells[1], g.Cells[2], g.Cells[3], g.Cells[4], g.Cells[5], g.Cells[6], g.Cells[7], g.Cells[8],
++---+---+---+`, resultIndexArray[0], resultIndexArray[1], resultIndexArray[2], resultIndexArray[3], resultIndexArray[4], resultIndexArray[5], resultIndexArray[6], resultIndexArray[7], resultIndexArray[8],
 	)
 }
 
--- a/15_tictactoe/go/grid/grid.go
+++ b/15_tictactoe/go/grid/grid.go
@@ -32,62 +32,63 @@ func (g Grid) String() string {
 // Create variable for empty cell, filled with X and with O
 // Return row of cells based on slice length, target index of slice and assign one of the 3 cell types
 // Return column of cells based and do the same as with the row
-func (g Grid) IsWinSimple() (bool, string) {
+func (g Grid) IsWin() (bool, string) {
 	// check horizontal
-	if g.Cells[0] == g.Cells[1] && g.Cells[0] == g.Cells[2] ||
-		g.Cells[3] == g.Cells[4] && g.Cells[3] == g.Cells[5] ||
-		g.Cells[6] == g.Cells[7] && g.Cells[6] == g.Cells[8] ||
+	if (g.Cells[0] == g.Cells[1] && g.Cells[0] == g.Cells[2]) ||
+		(g.Cells[3] == g.Cells[4] && g.Cells[3] == g.Cells[5]) ||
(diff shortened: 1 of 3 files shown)


riraum:
@nobe4 Tried to go further than this, but for now that's the easiest and quickest change to by a bit simpler.

nobe4:
@riraum you are comparing to the same value for all possible combinations. So you could do:

```go
if 
  (grid[0] == grid[1] && grid[0] == grid[2]) || // first line
  ...  // all the cases

  {
   // here, how would you determine who won?
}

return "no winner"

```

riraum:
@nobe4 Have it a bit more simplified, but it does still have many lines and based on your comment I should be make it a bit more DRY no?

nobe4 (commented):
You're right, I think without introducing another function, it will stay that complex.

I think you can move on for now.

However, if you want to make it neater, you can try to add a function like that:

```go
func (g Grid) checkLine(x,y,z int) (bool, string) {
	computerWin := "Computer wins!"
	humanWin := "Human wins!"

  	if g.Cells[x] != g.Cells[y] || g.Cells[x] != g.Cells[z] {
		return false, ""
	}

	if g.Cells[x] == "X" {
		return true, humanWin
	}
	
	return true, computerWin
}
```

Note that doing the early return and using `||` is faster than `&&` because a single `true` is enough to make the condition true, so not all tests are done.

Then you can use it like:

```go

func (g Grid) IsWin() (bool, string){
    if win, winner := checkLine(0, 1, 2); win {
      return win, winner
    }

    if win, winner := checkLine(3, 4 ,5); win {
      return win, winner
    }

    // ...

    return false, "no winner"
}
```

Once you have this, you could even put all the triplets (0,1,2), (3,4,5), ... into an array and loop over it. But that might be overdoing it haha

riraum:
@nobe4 Added, looks much neater and more pleasing.

Currently it does incorrectly show "no winner" when the user wins in the last round.

I think maybe I should first merge #30 after it's approved, then merge the master into this PR and clear up the conflicts and then finally do another PR for the final small touches, no?

15_tictactoe/go/grid/grid.go
```diff
@@ -18,78 +17,43 @@ func EmptyGrid() Grid {
...
+// Create variable for empty cell, filled with X and with O
+// Return row of cells based on slice length, target index of slice and assign one of the 3 cell types
+// Return column of cells based and do the same as with the row
+func (g Grid) IsWin() (bool, string) {
+	// check horizontal
+	g.checkLine(0, 1, 2)
+	//  check vertical
+	g.checkLine(0, 3, 6)
+	// check diagonally
+	g.checkLine(0, 4, 8)
```
nobe4: You will need to do a return here if one of the value returned is `true`

nobe4:
I think #30 should go first yeah :)

15_tictactoe/go/grid/grid.go
riraum: Done!

riraum: `; win` ensures that only the `bool` return gets evaluated?
It's the one part that seems new to me.

nobe4: Ah yes, see https://gobyexample.com/if-else and https://go.dev/tour/flowcontrol/6

nobe4: Also, note that those 3 tests aren't enough, you need all lines, columns and diagonals checked.

riraum:
@nobe4 Had some disagreements with Git, but prevailed and improved testing. Don't think I need to test the `EmptyGrid` function.

15_tictactoe/go/grid/grid.go
```diff
@@ -42,53 +40,36 @@ func (g Grid) String() string {
...
-	if g.Cells[0] == "O" && g.Cells[3] == "O" && g.Cells[6] == "O" ||
-		// 2nd row top to bottom vertically
-		g.Cells[1] == "O" && g.Cells[4] == "O" && g.Cells[7] == "O" ||
-		// 3rd row top to bottom vertically
-		g.Cells[2] == "O" && g.Cells[5] == "O" && g.Cells[8] == "O" {
-		return true, "Computer Won"
+// Create variable for empty cell, filled with X and with O
+// Return row of cells based on slice length, target index of slice and assign one of the 3 cell types
+// Return column of cells based and do the same as with the row
+func (g Grid) IsWin() (bool, string) {
```
nobe4, line 69: You are still missing a lot of test cases here.

Look at my addition in the tests.

15_tictactoe/go/grid/grid_test.go
```diff
@@ -35,14 +37,36 @@ func TestIsWin(t *testing.T) {
...
-			wantWinner: ""},
+			wantWinner: computerWin,
+		},
+		{
+			g: Grid{
+				Cells: [9]string{"O", " ", "X", "O", " ", "X", "O", "X", " "},
+			},
+			wantIsWin:  true,
+			wantWinner: computerWin,
+		},
```
nobe4: I don't think this new test passes

```suggestion
		{
			g: Grid{
				Cells: [9]string{"O", " ", "X", "O", " ", "X", "O", "X", " "},
			},
			wantIsWin:  true,
			wantWinner: computerWin,
		},
		{
			g: Grid{
				Cells: [9]string{
				"O", " ", "X", 
				"O", "O", "X", 
				"X", "X", "O",
				},
			},
			wantIsWin:  true,
			wantWinner: computerWin,
		},
```

15_tictactoe/go/grid/grid.go
riraum: Added some.

15_tictactoe/go/grid/grid_test.go
riraum: Added some and have high test coverage again.

nobe4 (approved):

PR #34, opened 2025-01-21
15_tictactoe - Refactored Play function to simplify

riraum:
#28 

> game.Play, refactor the game loop to simplify it
Hint: the loop should look like:

### DIFF
--- a/15_tictactoe/go/game/game.go
+++ b/15_tictactoe/go/game/game.go
@@ -29,8 +29,7 @@ Check if there is a winner, if yes, exit, if no, continue from the start
 func Play() {
 	// Initialize empty grid
 	g := grid.EmptyGrid()
-	i := 0
-	for {
+	for i := 0; ; i++ {
 		// Output empty grid to help understand the concept
 		// Ask player to select a grid index
 		fmt.Println(g.String())
@@ -49,7 +48,6 @@ func Play() {
 			// Add computer marker to correct index of grid
 			g.Cells[computerMove] = "O"
 		}
-		i++
 		//  Break condition in case of winner
 		checkWin, _ := g.IsWin()
 		if checkWin {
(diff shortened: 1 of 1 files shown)


riraum:
@nobe4 

> ` if i % 2 == 0 {
    // player make a move
  } else {
    // computer make a move
  }`

Why should I use this logic?
What does it protect me from?

Thinking of removing `i <= 9` and just let it loop until a winner is reached, what do you think? Don't think it serves a purpose with the `isValid` function already causing a forever loop with a wrong input.

nobe4:
> Thinking of removing i <= 9 and just let it loop until a winner is reached

good idea

> Why should I use this logic?
> What does it protect me from?

it could simplify the code further, since some of the logic is shared

riraum:
> it could simplify the code further, since some of the logic is shared

Could you expand on what you mean please? I'm not getting it yet.

nobe4:
> I'm not getting it yet.

Then nevermind, it's not that important.

15_tictactoe/go/game/game.go
```diff
@@ -29,61 +29,39 @@ Check if there is a winner, if yes, exit, if no, continue from the start
 func Play() {
 	// Initialize empty grid
 	g := grid.EmptyGrid()
-	fmt.Println(g.String())
-
-	for i := 0; i <= 9; i++ {
-		// Output empty grid to help understand the concept?
+	i := 0
+	for {
```
nobe4: You could put the initialization of i in the `for` line and skip the break condition:
```suggestion
	for i := 0 ;; i++{
```

nobe4 (approved):

15_tictactoe/go/game/game.go
riraum: Amazing, thank you!
I was wondering if I could use this syntax and not set a break condition initially.
Done!

PR #39, opened 2025-01-24
15_wordle - Check input against words

riraum:
- Move from hard coded word check, to check against list of valid words.


riraum:
No review needed yet 😬 Didn't knew adding you would send out a request immediately.

riraum:
@nobe4 
Should I move my `bytes.Equal` into a separate function? Trying to figure out if I have to, but I guess it makes it easier at least to check with a `bool` only...

16_wordle/word_list/word_list.go
nobe4, line 1: Please remove this file.

16_wordle/main.go
```diff
@@ -1,21 +1,133 @@
...
+	// Read word list
+	data, err := os.ReadFile("word_list/word_list")
+	if err != nil {
+		log.Fatal(err)
+	}
+	// os.Stdout.Write(data)
+	// debug
+	// fmt.Print(data)
+	// separate bytes
+	formattedData := bytes.Split([]byte(data), []byte("\n"))
```
nobe4: This is not "formatted" as much as it's "line-by-line" wordList.
A better name would be "wordList", "words", ...

16_wordle/word_list/word_list.go
riraum: Done

16_wordle/main.go
riraum: Done

16_wordle/main.go
```diff
@@ -1,21 +1,133 @@
...
+// 		if bytes.Equal(value, input) {
+// 			fmt.Print("Correct word!\n")
+// 			// if valid(input, word_list.wordSlice) {
+// 			return true
+// 		} else {
+// 			fmt.Print("Not correct word, try again!\n")
+// 			return false
+// 		}
+// 	}
+// }
```
nobe4: This is a good start. Maybe you can do

```go
func isValid(word []byte, validWords [][]byte) bool {
  // TODO
}
```

and then

```go
func isSecretWord(word, secret []byte) bool {
  // TODO
}
```

nobe4: If you want to move that to functions like

```go
type List struct {
  words [][]byte
  secretIndex int
}

func (l List) Valid(word []byte) bool{
  // ...
}

func (l List) Secret(word []byte) bool{
  // ...
}

```

you could as well, but the logic should be quite similar

nobe4 (commented):
It seems that you're trying to do too many things at once.

Take a step back, list all the functions that you need, make them as small as possible, implement them, add tests, and build from there.

16_wordle/main.go
nobe4: The difference between "valid" and "secret" is not great, maybe "valid" and "chosen"?
That's just naming, but naming is important https://nobe4.fr/posts/stars/#comment-via-types-and-names

16_wordle/main.go
```diff
@@ -1,21 +1,141 @@
...
+	// debug
+	// fmt.Print(data)
+	// separate bytes
+	words := bytes.Split([]byte(data), []byte("\n"))
+	// debug
+	// fmt.Printf("%q\n", formattedData)
+	// fmt.Printf("%q\n", bytes.Split([]byte(data), []byte("\n")))
+	// fmt.Println(bytes.Contains([]byte("brown"), formattedData))
 	for i := 0; ; i++ {
 		fmt.Printf("Enter 5 digit word\n>")
```
```suggestion
		fmt.Printf("Enter a 5 letters word\n>")
```

riraum: Done

riraum:
@nobe4 How about like this for this PR? 
Time for another PR?
Does the check against the list, added a hard coded word to check against specific word too.

16_wordle/main.go
```diff
@@ -1,21 +1,71 @@
...
+	return false
+}
+
+func isChosen(word, chosen []byte) bool {
+	if bytes.Equal(word, chosen) {
+		fmt.Print("Chosen word, yay!\n")
+		return true
 	}
+	return false
 }
```
nobe4: Those look good, can you add tests?

16_wordle/main.go
```diff
@@ -1,21 +1,71 @@
...
-			fmt.Print("Correct!\n")
+		// check if word on list
+		if isValid(input, words) && isChosen(input, chosenWordArray) {
 			break
-		} else {
-			fmt.Print("Not correct, try again!\n")
 		}
+		if isValid(input, words) && !isChosen(input, chosenWordArray) {
+			fmt.Print("But not chosen word, try again!\n")
+		}
```
nobe4: See if you can simplify this dual condition.

E.g. https://go.dev/play/p/iwJ4yG4Z8lZ

riraum: Hah, I had `continue` logic before. Tried again!

16_wordle/main.go
riraum: Added, not very DRY, but seems to work.
And only covers the check functions, not the main function.

16_wordle/main_test.go
```diff
@@ -0,0 +1,62 @@
...
+	"os"
+	"testing"
+)
+
+func TestIsValid(t *testing.T) {
+	data, err := os.ReadFile("word_list/word_list")
+	if err != nil {
+		log.Fatal(err)
+	}
+	words := bytes.Split([]byte(data), []byte("\n"))
```
nobe4: I wouldn't do that. In test, using the real data is not really worth the effort. Plus, here it's going to take a long time to read/split.

Instead, define a list of words yourself.

16_wordle/main_test.go
```diff
@@ -0,0 +1,62 @@
...
+
+	word1Str := "brown"
+	word1 := []byte(word1Str)
```
nobe4: This ought to be done in a single line.

16_wordle/main_test.go
```diff
@@ -0,0 +1,62 @@
...
+	word2Str := "aaaa"
+	word2 := []byte(word2Str)
+
+	tests := []struct {
+		word  []byte
+		words [][]byte
```
nobe4: If this never changes, I wouldn't put it in the test struct.

16_wordle/main_test.go
```diff
@@ -0,0 +1,62 @@
...
+	chosenWord := word1
+
+	tests := []struct {
+		word       []byte
+		chosenWord []byte
+		want       bool
+	}{
+		{word1, chosenWord, true},
+		{word2, chosenWord, false},
+	}
```
nobe4: You should move all the declaration in the tests here, since they change each time.

16_wordle/main_test.go
riraum: Done

16_wordle/main_test.go
riraum: Done

16_wordle/main_test.go
riraum: Done

16_wordle/main_test.go
riraum: What do you mean exactly? 
I removed `chosenWord` from the struct, but not sure it's that?

16_wordle/main.go
riraum: > If you want to move that to functions like
> 
> ```go
> type List struct {
>   words [][]byte
>   secretIndex int
> }
> 
> func (l List) Valid(word []byte) bool{
>   // ...
> }
> 
> func (l List) Secret(word []byte) bool{
>   // ...
> }
> ```
> 
> you could as well, but the logic should be quite similar

What is the benefit of defining a struct compared to my current state?
To be able to easily pick randomly an index as chosen word later on?

16_wordle/main_test.go
nobe4: no, what I meant was more something like that:

```
  tests := []struct {
    word       []byte
    chosenWord []byte
    want       bool
  }{
    { word: []byte("a"), chosenWord: []byte("b"), want: false },
    { word: []byte("a"), chosenWord: []byte("a"), want: true },
  }
```

16_wordle/main_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+	words := [][]byte{word1, word3}
+
+	tests := []struct {
+		word []byte
+		want bool
+	}{
+		{word1, true},
+		{word2, false},
+	}
```
nobe4: I would simplify it like so:

```suggestion
	words := [][]byte{
	  []byte{"brown"},
	  []byte{"aaaa"},
	}

	tests := []struct {
		word []byte
		want bool
	}{
		{[]byte("aaaa"), true},
		{[]byte("possa"), false},
	}
```

nobe4 (approved):
I'm just nitpicking at this moment, feel free to move forward.

16_wordle/main_test.go
riraum: I guess leads to better readability, even if I repeat myself a bit.
Done

16_wordle/main_test.go
riraum: Similar to the other suggestion. Done

PR #41, opened 2025-01-30
16_wordle - Add type scoped functions

riraum:
- Add type scoped functions


16_wordle/main.go
```diff
@@ -8,6 +8,10 @@ import (
 	"os"
 )
 
+type List struct {
+	words [][]byte
+}
```
riraum, line 13: @nobe4 in your example you added an int index.
Why would I need that?

Still have to change my function logic, but so far I can do without it.

nobe4: It was an idea on how to store the selection, but you don't have to do it. :)

16_wordle/main.go
```diff
@@ -34,6 +52,15 @@ func main() {
 	}
 }
 
+func (l List) IsValidList(word []byte) bool {
```
nobe4: Calling it `Contains` would be more inline with go std: https://pkg.go.dev/slices#Contains

riraum: Should I change the function logic therefore to `slices.Contains`, probably convert the `byte` values, instead of the current for range?

nobe4: Not necessarily. The implementation is fine as it is

riraum:
@nobe4 Like this?
Took a bit of reading, thinking and trying.
Not completely satisfied with the clarity of the variables yet...

I think the result is code that looks more complex, at least to me at this point.
Next step would be ofc to remove older code and fix the test again to work with the new logic.

16_wordle/main.go
```diff
@@ -61,22 +76,37 @@ func (l List) IsValidList(word []byte) bool {
...
+// }
 
-func isChosen(word, chosenWord []byte) bool {
+// func (l List) IsChosenWord(word []byte) bool {
+// 	return bytes.Equal(word, )
+// }
+
+func (l List) Chosen(word, chosenWord []byte) bool {
 	return bytes.Equal(word, chosenWord)
 }
```
nobe4: This is where storing the `chosen` word inside the struct would be useful. So you could do

```go
func (l List) Chosen(word []byte) bool {
  return bytes.Equal(l.chosen, word)
}
```

Otherwise, using the list here doesn't make much sense, as you're not using `l`

16_wordle/main.go
```diff
@@ -17,39 +21,92 @@ func main() {
...
+// 	return bytes.Equal(word, chosenWord)
+// }
+
+func (l List) Random(validWords [][]byte) []byte {
 	randomInt := rand.IntN(len(validWords))
 	// debug
-	fmt.Println(string(validWords[randomInt]))
+	// fmt.Println(validWords[randomInt])
 	return validWords[randomInt]
 }
```
nobe4: And here you could do:

```go
func (l List) Random() []byte {
  return l.words[rand.IntN(...)]
}
```

16_wordle/main.go
```diff
@@ -17,39 +21,92 @@ func main() {
...
+	// chosenWord := randomWord(validWordsStruct.words)
+	// debug
+	// fmt.Println(string(chosenWord))
 
 	for i := 0; ; i++ {
 		fmt.Printf("Enter 5 letter word\n>")
 		fmt.Scan(&input)
-		if isChosen(input, chosenWord) {
+
+		if validWordsStruct.Chosen(input, chosenWord) {
```
nobe4: This is just 
```suggestion
		if bytes.Equal(input, chosenWord) {
```

16_wordle/main.go
nobe4: But you also don't have to put _everything_ on `List`.

riraum: Hmm ya this function felt a bit pointless to write. 
Maybe I should re-consider changing the struct...

riraum: Alright, leaving it off for now :)

16_wordle/main.go
riraum: Done.
I had to specify the length, `(...)` would tell me that an operand was expected.

16_wordle/main.go
riraum: Done. Removed the Chosen function, seems not needed with this logic.

16_wordle/main.go
riraum: Then done!

16_wordle/main.go
```diff
@@ -16,40 +20,43 @@ func main() {
 		log.Fatal(err)
 	}
 
-	validWords := bytes.Split([]byte(rawWordList), []byte("\n"))
-	chosenWord := randomWord(validWords)
+	wordList := bytes.Split([]byte(rawWordList), []byte("\n"))
+
+	validWords := List{
+		words: wordList,
+	}
```
nobe4, line 27: how about adding another function to the `List` type:

```go
func NewList(path string) (List, error){
  rawWordList, err := os.ReadFile(path)
  // ...

  wordList := bytes.Split([]byte(rawWordList), []byte("\n"))

  // ...
}

```

https://go.dev/blog/error-handling-and-go

Errors are a very important part of the Go language, and understanding how they work is important :)

nobe4: BTW: feel free to add it in another PR

nobe4 (approved):

PR #42, opened 2025-01-31
16_wordle_check - Check characters

riraum:
- Check every input character, return if contained in `chosenWord`


riraum:
@nobe4 Accidentally pulled from the last PR, instead of merging...

Took me a while, but it seems to check every character now.
Time to change the logic to `byte` and add tests?

16_wordle/main.go
```diff
@@ -16,40 +22,80 @@ func main() {
...
+
+	for _, value := range inputSlice {
+		// debug
+		// fmt.Println("value print", value)
+		if slices.Contains(chosenWordSlice, value) {
+			containedSlice = append(containedSlice, value)
+		}
+	}
+	return containedSlice
 }
```
nobe4, line 101: This is technically valid, but I don't understand what you will do with the returned string.

In Wordle, you get a :green_square: for valid letters in correct places and :orange_square: for valid letters in incorrect places.
Since this function seems to be doing this, maybe instead of []string, you can return []int, with maybe 
- `0`: invalid
- `1`: valid, but wrong place
- `2`: valid, correct place

riraum: I wanted to start with a simpler output and then work on the more precise output in regards to position later on.

nobe4: gotcha :)

nobe4 (approved):

PR #43, opened 2025-02-01
16_wordle - Move create validWord List into function

riraum:
- Move list creation logic into function

### DIFF
--- a/16_wordle/main.go
+++ b/16_wordle/main.go
@@ -15,16 +15,18 @@ type List struct {
 func main() {
 	var input []byte
 
-	rawWordList, err := os.ReadFile("word_list/word_list")
-	if err != nil {
-		log.Fatal(err)
-	}
+	// rawWordList, err := os.ReadFile("word_list/word_list")
+	// if err != nil {
+	// 	log.Fatal(err)
+	// }
 
-	wordList := bytes.Split([]byte(rawWordList), []byte("\n"))
+	// wordList := bytes.Split([]byte(rawWordList), []byte("\n"))
 
-	validWords := List{
-		words: wordList,
-	}
+	// validWords := List{
+	// 	words: wordList,
+	// }
+
+	validWords, _ := NewList("word_list/word_list")
 
 	chosenWord := validWords.Random()
 	// debug
@@ -46,6 +48,21 @@ func main() {
 
 		fmt.Print("Invalid word, try again!\n")
 	}
+}
+
+func NewList(path string) (List, error) {
+	rawWordList, err := os.ReadFile(path)
+
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	wordList := bytes.Split([]byte(rawWordList), []byte("\n"))
+
+	validWords := List{
+		words: wordList,
+	}
+	return validWords, err
 }
 
 func (l List) Contains(word []byte) bool {
--- a/16_wordle/main.go
+++ b/16_wordle/main.go
@@ -50,7 +50,6 @@ func main() {
 	}
 }
 
-
 func NewList(path string) (List, error) {
 	rawWordList, err := os.ReadFile(path)
 
@@ -65,6 +64,7 @@ func NewList(path string) (List, error) {
 		words: wordList,
 	}
 	return validWords, nil
+}
 
 func formatInput(input []byte) string {
 	inputOutput := bytes.Join(bytes.Split(input, []byte("")), []byte(" "))
--- a/16_wordle/main.go
+++ b/16_wordle/main.go
@@ -15,17 +15,6 @@ type List struct {
 func main() {
 	var input []byte
 
-	// rawWordList, err := os.ReadFile("word_list/word_list")
-	// if err != nil {
-	// 	log.Fatal(err)
-	// }
-
-	// wordList := bytes.Split([]byte(rawWordList), []byte("\n"))
-
-	// validWords := List{
-	// 	words: wordList,
-	// }
-
 	validWords, _ := NewList("word_list/word_list")
 
 	chosenWord := validWords.Random()
--- a/16_wordle/main.go
+++ b/16_wordle/main.go
@@ -16,6 +16,8 @@ func main() {
 	var input []byte
 
 	validWords, _ := NewList("word_list/word_list")
+	// debug
+	fmt.Println(validWords)
 
 	chosenWord := validWords.Random()
 	// debug
--- a/16_wordle/main.go
+++ b/16_wordle/main.go
@@ -18,8 +18,6 @@ func main() {
 	var input []byte
 
 	validWords, _ := NewList("word_list/word_list")
-	// debug
-	fmt.Println(validWords)
 
 	chosenWord := validWords.Random()
 	fmt.Println("chosenWord:", string(chosenWord))
--- a/16_wordle/main.go
+++ b/16_wordle/main.go
@@ -44,6 +44,7 @@ func NewList(path string) (List, error) {
 
 	if err != nil {
 		log.Fatal(err)
+		return List{}, err
 	}
 
 	wordList := bytes.Split([]byte(rawWordList), []byte("\n"))
@@ -51,7 +52,7 @@ func NewList(path string) (List, error) {
 	validWords := List{
 		words: wordList,
 	}
-	return validWords, err
+	return validWords, nil
 }
 
 func (l List) Contains(word []byte) bool {
--- a/16_wordle/main_test.go
+++ b/16_wordle/main_test.go
@@ -3,6 +3,7 @@ package main
 import (
 	"log"
 	"os"
+	"path/filepath"
 	"reflect"
 	"testing"
 )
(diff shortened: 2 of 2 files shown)


16_wordle/main.go
```diff
@@ -48,6 +39,21 @@ func main() {
 	}
 }
 
+func NewList(path string) (List, error) {
+	rawWordList, err := os.ReadFile(path)
+
+	if err != nil {
+		log.Fatal(err)
+	}
```
nobe4: you might want to return the error here

16_wordle/main.go
```diff
@@ -48,6 +39,21 @@ func main() {
...
+
+	wordList := bytes.Split([]byte(rawWordList), []byte("\n"))
+
+	validWords := List{
+		words: wordList,
+	}
+	return validWords, err
```
nobe4: here, the err will always be `nil`, so you can also do

```suggestion
	return validWords, nil
```

16_wordle/main.go
riraum: I'll read about error handling. But added what I assume makes sense, since I do have return a `List` and can't just return the error I think.

16_wordle/main.go
riraum: Added

16_wordle/main.go
```diff
@@ -48,6 +39,22 @@ func main() {
...
+		return List{}, err
+	}
+
+	wordList := bytes.Split([]byte(rawWordList), []byte("\n"))
+
+	validWords := List{
+		words: wordList,
+	}
+	return validWords, nil
+}
```
nobe4, line 65: Looks nice!

If you want to write a test for it, have a look at https://pkg.go.dev/testing#T.TempDir
With this you can create a temporary dir, `os.Write` (or not!) in it and check the result.

nobe4 (approved):

16_wordle/main.go
riraum: Added a very non elegant, but seemingly working logic. Is that what you meant?

riraum: I'm trying.
I should start a file that has []byte words in it, to then run `NewList` and check if that results in a correctly formatted `wordList`.
But somehow I can't come up with the right way to test it.

nobe4: basically
```
write temporary file
create expected result
load the file
compare the two
```

nobe4: it's pretty much what you did already

16_wordle/main_test.go
```diff
@@ -35,3 +37,33 @@ func TestContains(t *testing.T) {
...
+		log.Fatal(err)
+	}
+
+	got, _ := NewList("dir/list")
+
+	if got != List{
+				words: [][]byte{
+					[]byte("brown", []byte("rossa")),
+		{
+			t.Errorf("NewList = %v", got)
```
nobe4: Make a new variable for your [][]byte list of words called "want", and use it everywhere it's needed.

```suggestion
			t.Errorf("want: %v but got: %v", got)
```

riraum: Done

16_wordle/main.go
riraum: Closer now, but having issues with the file, dir not existing :(

16_wordle/main_test.go
```diff
@@ -40,20 +41,22 @@ func TestContains(t *testing.T) {
 
 func TestNewList(t *testing.T) {
 	// tmpDir := t.TempDir()
-	err := os.WriteFile("t.TempDir(dir/list)", []byte("brown"), []byte("rossa"), 0666)
+	err := os.WriteFile("t.TempDir(dir/list)", []byte("brown"), 0666)
```
nobe4: This is incorrect, `t.TempDir()` returns a string which is a path to a temporary directory you can use to create temporary files there.

`os.WriteFile` expect its first argument to be a valid path to a file.

riraum: That is a bit of older code. Made some adjustments in later commits.

16_wordle/main_test.go
```diff
@@ -35,3 +39,32 @@ func TestContains(t *testing.T) {
...
+	// f, err := os.CreateTemp(tmpDir, "sample")
+
+	list := filepath.Join(tmpDir, "list")
+
+	err := os.WriteFile(list, []byte("brown"), 0666)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	got, _ := NewList("tmpDir/list")
```
nobe4:
```suggestion
	got, _ := NewList(list)
```

nobe4: Tho `listPath` would be a better name.

riraum: Wow, missed that 🤣

riraum: That was it, damn.

16_wordle/main_test.go
```diff
@@ -37,6 +41,29 @@ func TestContains(t *testing.T) {
...
+		words: [][]byte{
+			[]byte("brown"), []byte("rossa"), []byte("cuppy"),
+		},
+	}
+
+	if !reflect.DeepEqual(got.words, want.words) {
+		t.Errorf("want: %v, %v and %v but got: %v, %v and %v", string(want.words[0]), string(want.words[1]), string(want.words[2]), string(got.words[0]), string(got.words[1]), string(got.words[2]))
+	}
+}
+
```
riraum: @nobe4 Did a hacky string output for the 3 test words, as the byte version is not so readable for humans. 
Maybe should do this a bit better or okay like this?

nobe4: You can use `%s` to render bytes better:
https://go.dev/play/p/o7ageW4habl

16_wordle/main_test.go
```diff
@@ -37,6 +41,29 @@ func TestContains(t *testing.T) {
...
+	want := List{
+		words: [][]byte{
+			[]byte("brown"), []byte("rossa"), []byte("cuppy"),
+		},
+	}
+
+	if !reflect.DeepEqual(got.words, want.words) {
+		t.Errorf("want: %v, %v and %v but got: %v, %v and %v", string(want.words[0]), string(want.words[1]), string(want.words[2]), string(got.words[0]), string(got.words[1]), string(got.words[2]))
+	}
+}
```
nobe4: Instead of reflect (which is not recommended), you should use a comparison of length and then a loop.

riraum: Tried that and adding more tests in a struct.
Now it fails due to `expected 1 expressionsyntax` and I'm not successful in fixing it. Where can I look to get it working again?

16_wordle/main_test.go
```diff
@@ -37,6 +40,66 @@ func TestContains(t *testing.T) {
...
+	// 	want.words[index] != got.words[index]
+	//{
+	// 	t.Errorf("want: %s\n but got: %s\n", want.words[index], want.words[index]),
+	// }
+	// }
+
+	for index, test := range tests {
+		// if len(got.words) == len(test.want.words) {
+			if got.words[index] == test.want.words[index] {
+				t.Errorf("want: %s\n but got: %s\n", test.want.words[index], got.words[index]),
```
nobe4:
```suggestion
				t.Errorf("want: %s\n but got: %s\n", test.want.words[index], got.words[index])
```

16_wordle/main_test.go
```diff
@@ -37,6 +40,66 @@ func TestContains(t *testing.T) {
...
+	// 	continue
+	// }
+	// slices.Equal(got.words, want.words)
+	// for index := range got.words {
+	// 	want.words[index] != got.words[index]
+	//{
+	// 	t.Errorf("want: %s\n but got: %s\n", want.words[index], want.words[index]),
+	// }
+	// }
+
```
nobe4, line 76: here you should compare the length.

16_wordle/main_test.go
nobe4: See the new comments, you have a syntax issue.

16_wordle/main_test.go
riraum: Added

16_wordle/main_test.go
riraum: Done

16_wordle/main_test.go
```diff
@@ -37,6 +40,52 @@ func TestContains(t *testing.T) {
...
+	}
+
+	for index, test := range tests {
+		t.Run("", func(t *testing.T) {
+			tmpDir := t.TempDir()
+			listPath := filepath.Join(tmpDir, "listPath")
+
+			err := os.WriteFile(listPath, []byte(test.content), 0666)
+			if err != nil {
+				log.Fatal(err)
```
nobe4:
```suggestion
				t.Fatal(err)
```
 So that the test fails if it can't create the file

nobe4 (approved):

16_wordle/main_test.go
riraum: Done

PR #44, opened 2025-02-02
16_wordle - Add checkCar int output

riraum:
- Check if input characters are contained in `chosenWord` and return an indicator of yes and correct place or yes but incorrect place or not contained.

### DIFF
--- a/16_wordle/main.go
+++ b/16_wordle/main.go
@@ -87,18 +87,18 @@ func (l List) Random() []byte {
 	- Return slice of contained characters
 */
 
-func checkChar(chosenWordSlice []string, inputSlice []string) []string {
-	var containedSlice []string
+// func checkChar(chosenWordSlice []string, inputSlice []string) []string {
+// 	var containedSlice []string
 
-	for _, value := range inputSlice {
-		// debug
-		// fmt.Println("value print", value)
-		if slices.Contains(chosenWordSlice, value) {
-			containedSlice = append(containedSlice, value)
-		}
-	}
-	return containedSlice
-}
+// 	for _, value := range inputSlice {
+// 		// debug
+// 		// fmt.Println("value print", value)
+// 		if slices.Contains(chosenWordSlice, value) {
+// 			containedSlice = append(containedSlice, value)
+// 		}
+// 	}
+// 	return containedSlice
+// }
 
 func checkCharPrecise(chosenWordSlice []string, inputSlice []string) []int {
 	var containedSlice []int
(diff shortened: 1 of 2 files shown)


riraum:
@nobe4 Current attempt to give feedback if character is contained in `chosenWord` and if yes, if at the correct position.
Can restructure to `int` output, just found it easier to write like this.
Wondering if I should create a struct to output both characters and the corresponding `int` value.

16_wordle/main.go
```diff
@@ -99,3 +99,46 @@ func checkChar(chosenWordSlice []string, inputSlice []string) []string {
...
+func checkCharPrecise(chosenWordSlice []string, inputSlice []string) []string {
+	var containedSlice []string
+	// 1 = contained but incorrect loc
+	// 2 = contained in correc loc
+	// 0 not contained
+
+	// correct word check
+	if slices.Equal(chosenWordSlice, inputSlice) {
+		return []string{"Correct word!"}
+	}
```
nobe4: This is inconsistent with the expected output.
If you wanted to keep []string, you could do:

```suggestion
	if slices.Equal(chosenWordSlice, inputSlice) {
		return []string{"2","2","2","2","2"}
	}
```

16_wordle/main.go
```diff
@@ -99,3 +99,46 @@ func checkChar(chosenWordSlice []string, inputSlice []string) []string {
...
+
+	for iInput, inputValue := range inputSlice {
```
nobe4, line 114:
```suggestion
	for inputIndex, inputValue := range inputSlice {
```

16_wordle/main.go
```diff
@@ -99,3 +99,46 @@ func checkChar(chosenWordSlice []string, inputSlice []string) []string {
...
+		// if inputValue == chosenValue {
+		// 	// valueInt, _ := strconv.Atoi(inputValue)
+		// 	containedSlice = append(containedSlice, 2)
+		// }
+		for iChosen, chosenValue := range chosenWordSlice {
```
nobe4, line 119:
```suggestion
		for chosenIndex, chosenValue := range chosenWordSlice {
```

16_wordle/main.go
```diff
@@ -99,3 +99,46 @@ func checkChar(chosenWordSlice []string, inputSlice []string) []string {
 	}
 	return containedSlice
 }
+
+func checkCharPrecise(chosenWordSlice []string, inputSlice []string) []string {
```
nobe4: I really recommend using `[]int` instead, you will get confused later when it comes to compare the values inside the result.

Or, if you want to keep `[]string` for now, make sure that your return is consistent. Currently I list:
- `"Correct word!"`
- `"<char> inplace"`
- `"<char> outofplace"`

16_wordle/main.go
```diff
@@ -99,3 +99,46 @@ func checkChar(chosenWordSlice []string, inputSlice []string) []string {
...
+	for iInput, inputValue := range inputSlice {
+		// if inputValue == chosenValue {
+		// 	// valueInt, _ := strconv.Atoi(inputValue)
+		// 	containedSlice = append(containedSlice, 2)
+		// }
+		for iChosen, chosenValue := range chosenWordSlice {
+			if iInput == iChosen {
+				if inputValue == chosenValue {
+					// valueInt, _ := strconv.Atoi(inputValue)
+					containedSlice = append(containedSlice, inputValue+" inplace")
```
nobe4: Storing the `inputValue` here is useless, because you can get it from the index of `containedSlice`.

Another reason why storing `[]string` is making this more complex than you need it to.

16_wordle/main.go
nobe4: You almost have the full logic tho:

```go

for inIndex, inChar := range in {
  for chosenIndex, chosenChar := range chosen {
    if inChar != chosenChar{
      // TODO
    } else {
      if inIndex == chosenIndex {
        // TODO
      } else {
        // TODO
      }
    }
  }
}

```

nobe4: Maybe here would be a good idea to start with the tests, and to fit the implementation to the tests?

E.g.
```go

var tests := []struct {
  input string
  chosen string
  want []int
}{
  {
    input: "aaaaa",
    chosen: "aaaab",
    want: []int{2,2,2,2,0},
  },
  {
    input: "acaab",
    chosen: "aacab",
    want: []int{2,1,1,2,2},
  },
  // ...
}

```

Because once you have a list of the different cases, you can implement much faster to fit those cases.

riraum: Okay, went back to `[]int`

riraum:
@nobe4 Added tests. Tried various ways, based on your comment, writing it without else statements to have it more clear. 
As soon as I add a 3rd check to it it gives me a slice of more than 5 characters.

> You almost have the full logic tho:
> 
> ```go
> for inIndex, inChar := range in {
>   for chosenIndex, chosenChar := range chosen {
>     if inChar != chosenChar{
>       // TODO
>     } else {
>       if inIndex == chosenIndex {
>         // TODO
>       } else {
>         // TODO
>       }
>     }
>   }
> }
> ```

16_wordle/main.go
```diff
@@ -99,3 +99,88 @@ func checkChar(chosenWordSlice []string, inputSlice []string) []string {
...
+	//       if inIndex == chosenIndex {
+	//         // TODO
+	//       } else {
+	//         // TODO
+	//       }
+	//     }
+	//   }
+	// }
+
+	for inputIndex, inputValue := range inputSlice {
```
nobe4: Here you have 3 cases:

- 2: same value, same index, you can get that with `inputValue == chosenValue && inputIndex == chosenIndex`
- 1: the value is in the chosen word list
- 0: the value is not in the chosen word list

Maybe you don't need a double loop but there's a different way to approach it?

16_wordle/main.go
```diff
@@ -87,15 +88,80 @@ func (l List) Random() []byte {
...
+	// 	}
+	// }
+
+	if inputSlice[0] == chosenWordSlice[0] {
+		containedSlice = append(containedSlice, 2)
+	} else if slices.Contains(chosenWordSlice, inputSlice[0]) {
+		containedSlice = append(containedSlice, 1)
+	} else {
+		containedSlice = append(containedSlice, 0)
+	}
```
nobe4: this is actually the logic I would have implemented, can you use a loop now instead of copying / pasting this code many times?

16_wordle/main_test.go
```diff
@@ -35,3 +36,29 @@ func TestContains(t *testing.T) {
...
+	}{
+		{input: []string{"a", "a", "a", "a", "a"},
+			chosen: []string{"a", "a", "a", "a", "b"},
+			want:   []int{2, 2, 2, 2, 1},
+		},
+		{
+			input:  []string{"a", "c", "a", "a", "b"},
+			chosen: []string{"a", "a", "a", "b", "a"},
+			want:   []int{2, 0, 2, 1, 1},
+		},
```
nobe4: Also, when you have converted into a loop, you can test with input with different length, so it becomes easier to test for special cases

16_wordle/main.go
```diff
@@ -78,24 +79,34 @@ func (l List) Random() []byte {
...
-	- Add character that is contained to new slice
-	- Return slice of contained characters
+- Loop through each character
+- Add character that is contained to new slice
+- Return slice of contained characters
 */
+func checkChar(chosenWordSlice []string, inputSlice []string) []int {
+	var containedSlice []int
+	// 1 = contained but incorrect loc
+	// 2 = contained in correc loc
```
```suggestion
	// 2 = contained in correct loc
```

16_wordle/main.go
```diff
@@ -78,24 +79,34 @@ func (l List) Random() []byte {
 }
 
 /*
-	Check every character of `input` and return character(s) that are in `chosenWord`
+Check every character of `input` and return character(s) that are in `chosenWord`
 
-	Convert at some point between byte and string type to return human readable output
+# Convert at some point between byte and string type to return human readable output
```
nobe4, line 76: Maybe now you can undo the conversion?

```suggestion
Convert at some point between byte and string type to return human readable output
```

PR #45, opened 2025-02-05
16_wordle - Refine output to match reqs

riraum:
- Move towards output as in the specs.

### DIFF
--- a/16_wordle/main.go
+++ b/16_wordle/main.go
@@ -35,9 +35,8 @@ func main() {
 		fmt.Printf("Enter 5 letter word\n>")
 		fmt.Scan(&input)
 
-		// check characters
-		inputOutput := bytes.Split(input, []byte(""))
-		fmt.Println(string(bytes.Join(inputOutput[:], []byte(" "))))
+		inputOutput := bytes.Join(bytes.Split(input, []byte("")), []byte(" "))
+		fmt.Println(string(inputOutput))
 
 		getResult := checkChar(chosenWord, input)
 		coloredOutput := strings.Join(coloredResult(getResult)[:], "")
--- a/16_wordle/main.go
+++ b/16_wordle/main.go
@@ -34,6 +34,7 @@ func main() {
 	for i := 0; ; i++ {
 		fmt.Printf("Enter 5 letter word\n>")
 		fmt.Scan(&input)
+		fmt.Println(formatInput(input))
 
 		inputOutput := bytes.Join(bytes.Split(input, []byte("")), []byte(" "))
 		fmt.Println(string(inputOutput))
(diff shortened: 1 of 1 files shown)


riraum:
@nobe4 Please review. 
Considered adding the emoji to `checkChar` instead of a separate function, but left it separate for now.
Could check input length and do other validations too and more tests...

16_wordle/main.go
```diff
@@ -57,6 +57,20 @@ func main() {
...
+func coloredResult(checkResult []int) []string {
+	var result []string
+	for _, value := range checkResult {
+		if value == 2 {
+			result = append(result, "🟩")
+		} else if value == 1 {
+			result = append(result, "🟨")
+		} else {
+			result = append(result, "🟥")
+		}
```
nobe4: The logic is good here :+1:
But the conditions are a bit ugly :smile: 

Let's do something better here.

2 solutions:
- https://go.dev/wiki/Switch
- Using a `map[int]string` that maps the value to the emoji.

16_wordle/main.go
```diff
@@ -29,19 +29,19 @@ func main() {
...
 
 	for i := 0; ; i++ {
 		fmt.Printf("Enter 5 letter word\n>")
 		fmt.Scan(&input)
 
 		// check characters
-		fmt.Println(checkChar(chosenWord, input))
-		fmt.Println(chosenWordStrSlice)
+		inputOutput := bytes.Split(input, []byte(""))
+		fmt.Println(string(bytes.Join(inputOutput[:], []byte(" "))))
```
```suggestion
		fmt.Println(string(input))
```

riraum: I wanted to match the output from the README, where one emoji block is directly underneath the input character.
There's probably better ways to achieve that, but it's what I could think of. Had with multiple variables, lines which looked worse I think.

nobe4: Ah, I see, then I would suggest to move this into a function :)

And you can simplify with:

```suggestion
		inputOutput := bytes.Join(bytes.Split(input, []byte("")),[]byte(" "))
		fmt.Println(string(inputOutput))
```

16_wordle/main.go
riraum: Done with switch!

16_wordle/main.go
riraum: Done

16_wordle/main.go
```diff
@@ -57,6 +57,32 @@ func main() {
...
+				result = append(result, "🟩")
+			}
+		case 1:
+			{
+				result = append(result, "🟨")
+			}
+		default:
+			{
+				result = append(result, "🟥")
+			}
```
```suggestion
		case 2:
			result = append(result, "🟩")
		case 1:
			result = append(result, "🟨")
		default:
			result = append(result, "🟥")
```

nobe4 (approved):

16_wordle/main.go
nobe4: You might need to format this tho.

riraum: Done, also removed duplicate print statement.

nobe4 (approved):

Issue #48, opened 2025-02-08
Wordle QA


nobe4:
General guideline: if it's an edge-case or something you didn't foresee, add a test for this specific case.

- [x] if you input a word longer than 5, it fails:
  ```
  >abcdefg
  a b c d e f g
  panic: runtime error: index out of range [5] with length 5
  
  goroutine 1 [running]:
  main.checkChar({0xc000119b68, 0x5, 0xc000108ef0?}, {0xc000098008, 0x7, 0x0?})
  	/home/n/dev/riraum/forward/16_wordle/main.go:104 +0x210
  main.main()
  	/home/n/dev/riraum/forward/16_wordle/main.go:30 +0x1dd
  exit status 2
  ```
- [x] don't print the solution at the start of the game
- [x] don't print "Enter 5 letter word" on each input
- [x] don't run the scoring + coloring for invalid words

  ```
  >a
  a
  🟨
  Invalid word, try again
  >12345
  1 2 3 4 5
  🟥🟥🟥🟥🟥
  Invalid word, try again!
  Enter 5 letter word
  ```
- [x] display the number of attempts left, currently there are none

  E.g.
  ```
  1/5 >
  2/5 >
  3/5 >
  ...
  ```
- [x] [This code](https://github.com/riraum/forward/blob/3f84f3d47dda787f820750ae9e895b248e55279d/16_wordle/main.go#L50-L53) is incorrect, the `return` is never reached. See https://pkg.go.dev/log#Fatal. I recommend to return the error and handle it in the `main` function instead.
- [x] The extra [`[]byte(...)` cast](https://github.com/riraum/forward/blob/3f84f3d47dda787f820750ae9e895b248e55279d/16_wordle/main.go#L55), looks superfluous.
- [x] You can directly return [the valid word list](https://github.com/riraum/forward/blob/3f84f3d47dda787f820750ae9e895b248e55279d/16_wordle/main.go#L57-L60), no need for a temporary variable.
- [x] Add a test for the function `Contains` with _at least_ 2 words in the list, where the valid word is not the first word in the list. And see how your current implementation of `Contains` is wrong.
- [x] You should rewrite the `TestContains` function as well. Currently you do some weird stuff with `got1` and `got2`. Look at the other tests.

riraum:
Implemented.

PR #49, opened 2025-02-09
16_wordle - Refactor to implement feedback

riraum:
- Implement feedback from #48

### DIFF
--- a/16_wordle/main.go
+++ b/16_wordle/main.go
@@ -60,6 +60,9 @@ func main() {
 
 func NewList(path string) (List, error) {
 	rawWordList, err := os.ReadFile(path)
+	if err != nil {
+		return List{}, err
+	}
 	wordList := bytes.Split((rawWordList), []byte("\n"))
 
 	return List{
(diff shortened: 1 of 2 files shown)


16_wordle/main.go
```diff
@@ -46,18 +60,11 @@ func main() {
 
 func NewList(path string) (List, error) {
 	rawWordList, err := os.ReadFile(path)
```
nobe4, line 67: You should return the error as soon as possible in most cases. In this case it would be here :)

16_wordle/main.go
```diff
@@ -82,7 +89,9 @@ func coloredResult(checkResult []int) []string {
 
 func (l List) Contains(word []byte) bool {
 	for _, value := range l.words {
-		return bytes.Equal(word, value)
+		if bytes.Equal(word, value) {
```
nobe4, line 100: 👍

riraum: I think I broke it before, now I un-broke it again :)

16_wordle/main.go
riraum: Something like this?

nobe4: Yes :)

nobe4 (approved):

